PR #19, opened 2023-08-14
add models and update sh, readme

Jintao-Huang:
1. update readme, readme in example, readme_CN in example
2. update push_to_hub sh
3. add num_train_epochs in sh
4. add chatglm2-6b-32k
5. add lora_target_modules ALL
6. update is_dist


examples/pytorch/llm/scripts/qwen_7b/qlora_ddp/sft.sh
```diff
@@ -7,20 +7,30 @@ torchrun \
...
+    --warmup_ratio 0.03 \
     --eval_steps 50 \
     --save_steps 50 \
     --save_total_limit 2 \
     --logging_steps 10 \
     --use_flash_attn false \
+    --push_to_hub false \
+    --hub_model_id qwen-7b-qlora \
+    --hub_private_repo true \
+    --hub_token 'your-sdk-token' \
```
tastelikefeet, line 36: What if the token is not provided

tastelikefeet, line 36: What if the token is not provided

Jintao-Huang: If the token is None, we will get env variable MODELSCOPE_API_TOKEN. 
If the sdk token is wrong, an exception will is raise.

Jintao-Huang: If push_to_hub is False, the hub token is not used

tastelikefeet (approved):

PR #44, opened 2023-09-04
fix template bug2

Jintao-Huang:
1. add copyright
2. fix template bug


Jintao-Huang:
fix multi-node DDP bug

tastelikefeet (approved):

PR #112, opened 2023-10-12
Support mistral 7b

Jintao-Huang:
1. support openbuddy mistral 7b
2. support transformers>=4.34
3. add requires in MODEL_MAPPING
4. rename dataset and model
5. add sh


Jintao-Huang:
6. support mistral-7b, mistral-7b-chat

tastelikefeet (approved):

PR #227, opened 2023-12-17
fix dtype='fp16' sft bug

Jintao-Huang:


Jintao-Huang:
issue: https://github.com/modelscope/swift/issues/225

tastelikefeet (approved):

PR #354, opened 2024-01-30
Support xcomposer

Jintao-Huang:


tastelikefeet (approved):

zodiacg:
Can finetuning/inferring internlm-xcomposer2-vl be supported in the same way?

PR #472, opened 2024-02-29
Support LLaMAPRO and LoRA+

tastelikefeet:


Jintao-Huang (approved):

AU-CPU:
Hello, I have a question. When I selects model_type=llamapro, will freeze_llm=false take effect at this time? During the training process, llm will only fine-tune the newly added blocks as indicated in the llamapro paper, and freeze from The blocks inherited from the basic model . or will still follow freeze_llm=false and all parameters will follow the training?

PR #477, opened 2024-02-29
Supports to use custom datasets for models quantization

DavidQiuChao:
The previous quantization interface did not support calibration using custom datasets, so I added a feature that supports calibration using custom datasets. And I have introduced the usage method in the document.

### DIFF
--- a/swift/llm/utils/argument.py
+++ b/swift/llm/utils/argument.py
@@ -627,6 +627,11 @@ class ExportArguments(InferArguments):
     commit_message: str = 'update files'
 
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        register_custom_dataset(self)
+
+
 @dataclass
 class DPOArguments(SftArguments):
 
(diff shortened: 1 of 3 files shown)


swift/llm/utils/argument.py
```diff
@@ -627,6 +627,11 @@ class ExportArguments(InferArguments):
     commit_message: str = 'update files'
 
 
+    def __post_init__(self) -> None:
```
Jintao-Huang, line 630: Here, the dataset registration can be removed. It is not necessary to register the dataset again, as it will be registered in the InferArguments.

Jintao-Huang:
Hello, the main branch already supports quantization of custom datasets. You can refer to the LLM quantization documentation here: https://github.com/modelscope/swift/blob/main/docs/source/LLM/LLM%E9%87%8F%E5%8C%96%E6%96%87%E6%A1%A3.md

Jintao-Huang:
https://github.com/modelscope/swift/pull/484

PR #594, opened 2024-03-23
fix mix_dataset_sample float

Jintao-Huang:


Jintao-Huang:
fix issue: https://github.com/modelscope/swift/issues/593

PR #872, opened 2024-05-07
fix list index out of range bugs of model_name and model_author 

tiandiweizun:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information
fix list index out of range bugs of model_name and model_author


swift/llm/utils/dataset.py
```diff
@@ -1157,9 +1157,9 @@ def _preprocess_self_cognition_dataset(
     # model_name: Tuple[zh, en]
     assert model_name[0] is not None
     assert model_author[0] is not None
-    if model_name[1] is None:
+    if len(model_name) == 1:
```
Jintao-Huang: 要不 len(model_name) == 1 or model_name[1] is None

tiandiweizun: https://github.com/modelscope/swift/pull/885

PR #879, opened 2024-05-08
fix swift cli exit code if subprocess is failed

pitt-liang:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Job executed on remote machine relies on the exit code of the command to determine if the job is successful or a failure. Currently, the swift cli always returns an exit code of 0, which leads to a situation where even if the training job actually fails, it is still considered successful at the job level.

## Experiment results

before 
image

current PR branch test result:

image

### DIFF
--- a/swift/cli/main.py
+++ b/swift/cli/main.py
@@ -61,7 +61,9 @@ def cli_main() -> None:
     else:
         args = ['torchrun', *torchrun_args, file_path, *argv]
     print(f"run sh: `{' '.join(args)}`", flush=True)
-    subprocess.run(args)
+    result = subprocess.run(args)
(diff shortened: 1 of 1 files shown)


Jintao-Huang:
Hello~ Thank you for your code. Please pass the lint test.

https://github.com/modelscope/swift/blob/main/CONTRIBUTING_CN.md#%E4%BB%A3%E7%A0%81%E8%A7%84%E8%8C%83%E5%92%8C%E5%BC%80%E5%8F%91%E6%96%B9%E5%BC%8F

tastelikefeet (approved):

PR #1193, opened 2024-06-20
Fix dataset concatenation

tastelikefeet:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Dataset concatenation may raise following errors:

```text
_check_if_features_can_be_aligned
    raise ValueError(
ValueError: The features can't be aligned because the key history of features {'system': Value(dtype='null', id=None), 'history': Sequence(feature=Sequence(feature=Value(dtype='string', id=None), length=-1, id=None), length=-1, id=None), 'query': Value(dtype='string', id=None), 'response': Value(dtype='string', id=None)} has unexpected type - Sequence(feature=Sequence(feature=Value(dtype='string', id=None), length=-1, id=None), length=-1, id=None) (expected either Sequence(feature=Value(dtype='null', id=None), length=-1, id=None) or Value("null").
```

This is because some dataset has empty values and None values, and another one has normal history values, so the arrow_dataset will treat them as difference types.

How to solve:

reduce column after the dataset instantiated, and before the concatenation.

## Experiment results

### DIFF
--- a/swift/llm/utils/dataset.py
+++ b/swift/llm/utils/dataset.py
@@ -2262,6 +2262,7 @@ def get_dataset(
             if 'history' in ds.features:
                 features = ds.features.copy()
                 features['_history'] = Sequence(feature=Sequence(feature=Value(dtype='string')))
+            if 'system' in ds.features:
                 features['_system'] = Value(dtype='string')
             ds = ds.map(_reduce_column, load_from_cache_file=False, features=features)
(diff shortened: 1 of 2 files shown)


Jintao-Huang (approved):

zodiacg:
这个补丁没有发布吗？遇到这个问题了

PR #1334, opened 2024-07-09
fix florence model

hjh0119:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information
Add length truncation for Florence models: https://github.com/modelscope/swift/issues/1267
Add the InternVL2-40B model


lucasjinreal:
Hi, I think force the lenght of attention mask and input ids might not be a good way.

We need passing the training samples maybe? if the content length extremly large, it's better drop it than train on a trucated one

Jintao-Huang (approved):

lucasjinreal:
Can we set max lenght to 4096?

hjh0119:
> Can we set max lenght to 4096?

You can manually set the parameter `--max_length`, but unexpected results may occur because the model itself does not support very long lengths.

lucasjinreal:
I found I still got same error ,when set max-length to 2048.
WHy

lucasjinreal:
It's really can not exceed 1024, limit to 1024 then everything is OK, why???

PR #1784, opened 2024-08-21
Add Internvl2TemplateWithAngles template 

DYF-AI:
# PR type

# PR information

Template A is added to support entering image angles，and the dataset format is as follows:

`{"query": "please describe the image.", "response": "An old man combing the hair of a young boy.", "images": ["/mnt/g/dongyongfei786/custom-swift/examples/data_processing/output/train/235812.jpg"], "angles": [0]}`

add custom templates：
`
--template internvl2-angle
`


Jintao-Huang:
Thank you very much for your PR. Could you please describe the reason for submitting it?😊

DYF-AI:
> Thank you very much for your PR. Could you please describe the reason for submitting it?😊

Hello, I'm glad to receive your reply. The reason for this PR is that when extracting multimodal image information, the usual practice is to classify the images into angles (such as 0, 90, 180, 270, etc.), and then input the corrected angle images into the model. Based on our experience, this approach works better with a small number of samples for fine-tuning training. The original data format did not support the input of image angles, so we are submitting this PR.

PR #1804, opened 2024-08-23
fix megatron_patch_path

wning13:
# PR type
- [x] Bug Fix

# PR information

When the environment variable `PAI_MEGATRON_PATCH_PATH` is set, the following error occurs: 
```
UnboundLocalError: local variable 'megatron_patch_path' referenced before assignment.
```

This error occurs because the variable `megatron_patch_path` is not declared in this context.

The change made was to assign the value of the environment variable `PAI_MEGATRON_PATCH_PATH` to `megatron_patch_path`.

## Experiment results


Jintao-Huang:
thanks

Jintao-Huang (approved):

PR #2100, opened 2024-09-23
Fix the issue with media_offset in owl3 when batch_size > 1.

LukeForeverYoung:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

When the batch size > 1, the media_offset in mPLUG-Owl3 needs to be padded. We use the last value of each media_offset to fill the padding. Additionally, note that the images in the same batch are stacked in order, so the value of media_offset should be incremented by the number of images preceding it


Jintao-Huang (approved):

Jintao-Huang:
Thank you very much.

Issue #2128, opened 2024-09-25
用全量微调过的llava-onevision-qwen2-0_5b-ov权重在chartqa数据集上进行eval遇到RuntimeError: Failed to import swift.llm.deploy because of the following error (look up to see its traceback): cannot import name 'add_vllm_request' from 'swift.llm.utils' (/data/swift/swift/llm/utils/__init__.py)问题


Psycho-9:
用经过full stf之后的权重在chartqa和docvqa等进行验证，命令是swift eval --model_type llava-onevision-qwen2-0_5b-ov --local_repo_path /data/swift/LLaVA/ --ckpt_dir /data/swift/llava-onevision-qwen2-0_5b-ov/checkpoint-20200/ --eval_dataset ChartQA_TEST --eval_output_dir eval_outputs/llava-onevision-qwen2-0_5b-ov-chartqa `，但是发生下面错误
```
Traceback (most recent call last):
  File "/data/qdluo/anaconda3/envs/swift/lib/python3.10/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/data/qdluo/anaconda3/envs/swift/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/data/swift/swift/llmTraceback (most recent call last):
  File "/data/qdluo/anaconda3/envs/swift/lib/python3.10/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/data/qdluo/anaconda3/envs/swift/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/data/swift/swift/llm/eval.py", line 140, in run_custom_model
    from swift.llm import deploy_main
  File "<frozen importlib._bootstrap>", line 1075, in _handle_fromlist
  File "/data/swift/swift/utils/import_utils.py", line 54, in __getattr__
    module = self._get_module(self._class_to_module[name])
  File "/data/swift/swift/utils/import_utils.py", line 66, in _get_module
    raise RuntimeError(
RuntimeError: Failed to import swift.llm.deploy because of the following error (look up to see its traceback):
cannot import name 'add_vllm_request' from 'swift.llm.utils' (/data/swift/swift/llm/utils/__init__.py)
[ERROR:swift] Cannot get model_type from the deploy service, please check the error to continue eval
[INFO:swift] The deployment process has been terminated.
Traceback (most recent call last):
[...]

PR #2141, opened 2024-09-26
fix deploy vllm

Jintao-Huang:
# PR type
- [x] Bug Fix


Jintao-Huang:
FIX https://github.com/modelscope/ms-swift/issues/2128

PR #2754, opened 2024-12-24
add 'right' option for 'truncation_strategy'

zsxm1998:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates

# PR information

A new 'right' parameter has been added to the truncation_strategy, corresponding to right-side truncation. This is necessary, especially in the training of multimodal large models. For instance, if an image is input at the beginning of the data but the model's output exceeds the max_length, left-side truncation alone may result in the image being cut off, causing the model to lose crucial context for generating the result.

## Experiment results

No experiment.

### DIFF
--- a/docs/source/Instruction/命令行参数.md
+++ b/docs/source/Instruction/命令行参数.md
@@ -45,7 +45,7 @@
 - 🔥template: 对话模板类型，默认使用model对应的template类型。`swift pt`会将对话模版转为生成模板使用
 - 🔥system: 自定义system字段，默认为None，使用template的默认system
 - 🔥max_length: 单样本的tokens最大长度，默认为None，不做限制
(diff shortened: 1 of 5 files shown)


Jintao-Huang:
hello, thanks for your PR.

The `delete` strategy seems to be able to handle this situation. If a right trim results in an entirely clean response, I believe it will throw an error.

zsxm1998:
Thank you for your thoughtful response. 

However, I would like to clarify that the delete strategy directly removes overly long samples, which may lead to wasted samples. Additionally, I apologize for any confusion, but I did not fully understand your last sentence. Would you be so kind as to provide a more detailed explanation? I greatly appreciate your insights and look forward to further discussion.

Jintao-Huang (approved):

Jintao-Huang:
I understand. thanks

PR #3135, opened 2025-02-17
fix grpo pt infer generation_config

Jintao-Huang:


tastelikefeet (approved):

Jintao-Huang:
This bug will cause the temperature passed in GRPO training to become ineffective when `do_sample` in the model's `generation_config` is set to False.

PR #3350, opened 2025-03-03
support phi4-multimodal

Jintao-Huang:


Jintao-Huang:
training shell: https://github.com/modelscope/ms-swift/blob/main/examples/train/plugins/tuner.sh

PR #3458, opened 2025-03-12
fix seed

Jintao-Huang:


Jintao-Huang:
squash from https://github.com/modelscope/ms-swift/pull/3457

PR #3460, opened 2025-03-12
Resume sample

BC-A:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

Currently, when sampling, if the sampling process fails for some reason, then sampling needs to be started from the beginning, which greatly wastes inference resources. Therefore, a resume function is added to persist the data to disk by batch and record the persisted index. If the process fails, then when resuming, the index is read and data smaller than the index will not be sampled, thus achieving the recovery function.

## Experiment results

index_resume indicates the persisted index. When resuming, the sample will start from index_resume+1.

image


tastelikefeet:
Can you run `pip install pre-commit` then `pre-commit run --all-files` to lint the code please?

tastelikefeet (approved):

PR #3692, opened 2025-03-27
Grpo vl72b script

hjh0119:
# PR type
- [ ] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Experiment results

### DIFF
--- a/examples/train/grpo/lora_qwenvl72b.sh
+++ b/examples/train/grpo/lora_qwenvl72b.sh
@@ -20,8 +20,8 @@ swift rlhf \
   --per_device_train_batch_size 1 \
   --per_device_eval_batch_size 1 \
   --gradient_accumulation_steps 1 \
-  --eval_steps 1000 \
-  --save_steps 2000 \
+  --eval_steps 500 \
+  --save_steps 500 \
   --learning_rate 1e-6 \
   --save_total_limit 2 \
   --logging_steps 1 \
--- a/examples/train/grpo/lora_qwenvl72b.sh
+++ b/examples/train/grpo/lora_qwenvl72b.sh
@@ -27,7 +27,6 @@ swift rlhf \
   --warmup_ratio 0.05 \
   --dataloader_num_workers 4 \
   --max_completion_length 2048 \
-  --reward_funcs format \
   --num_generations 8 \
   --use_vllm true \
   --vllm_gpu_memory_utilization 0.5 \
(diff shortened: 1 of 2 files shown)


tastelikefeet (approved):

sys-reasoner:
Hi guys, 

In this shell command what does "--move_model_batches 40" actually mean? What is this parameter related to?
For example, when expanding to a node with 16 gpus, or even multi nodes, should this parameter be modified?

Jintao-Huang (approved):

hjh0119:
@sys-reasoner The motivation behind this approach is as follows: after each update to the model weights, we need to synchronize these weights with vLLM. In the zero3 setting, the model weights are distributed across different devices, which necessitates gathering these weights first. However, this gathering step often leads to out-of-memory (OOM) issues.

A natural solution to this problem is to group the model weights, allowing us to gather and synchronize them in batches. This strategy alleviates the memory overhead incurred during the weight gathering process. The `move_model_batches` 
parameter specifies the number of groups into which the model weights will be divided.

For MLLM, we only split the LLM portion, as it constitutes the majority of the model's weights. The choice of 40 as the hyperparameter stems from the fact that the LLM part of the Qwen2.5-VL-72B model consists of 80 layers, making 40 a suitable choice for partitioning.

hjh0119:
image

sys-reasoner:
> @sys-reasoner The motivation behind this approach is as follows: after each update to the model weights, we need to synchronize these weights with vLLM. In the zero3 setting, the model weights are distributed across different devices, which necessitates gathering these weights first. However, this gathering step often leads to out-of-memory (OOM) issues.
> 
> A natural solution to this problem is to group the model weights, allowing us to gather and synchronize them in batches. This strategy alleviates the memory overhead incurred during the weight gathering process. The `move_model_batches` parameter specifies the number of groups into which the model weights will be divided.
> 
> For MLLM, we only split the LLM portion, as it constitutes the majority of the model's weights. The choice of 40 as the hyperparameter stems from the fact that the LLM part of the Qwen2.5-VL-72B model consists of 80 layers, making 40 a suitable choice for partitioning.

wow this approach is very promising, well done! So based on your comment, the hyperparam "move_model_batches" is only related to model's layer number itself, which is  divisible by "move_model_batches"? Do I need to modify when scaling the gpus or gpu nodes?

hjh0119:
>  the hyperparam "move_model_batches" is only related to model's layer number itself, which is divisible by "move_model_batches"? Do I need to modify when scaling the gpus or gpu nodes?

1. Yes, but it is not necessarily divisible; the implementation in the code uses ceil.
2. No. However, as the number of GPUs increases, zeroing will reduce GPU memory usage, so you might consider lowering the "move_model_batches" parameter to accelerate the process, as this is a time-for-space trick.

dwang-sflscientific:
请问我如果需要训练自己的mllm grpo, 是不是准备好类似lmms-lab/multimodal-open-r1-8k-verified的数据集就好？

PR #3778, opened 2025-04-06
Error message fix - truncation strategy for RLHF

Reichenbachian:
# PR type
- [X] Bug Fix

# PR information

If you truncate from the left (overwriting whatever the user actually supplied from the CLI), it is awfully confusing. I ended up debugging this for a while when my multimodal model had its image tokens cropped, and I got a VLLM error saying I had an incorrect number of image tokens. I figure instead of overwriting the user preference, raising here is probably better.

Not really sure why we don't just respect the user argument preference instead, and happy to allow that too.

### DIFF
--- a/swift/llm/argument/rlhf_args.py
+++ b/swift/llm/argument/rlhf_args.py
@@ -139,7 +139,8 @@ class RLHFArguments(GRPOArguments, PPOArguments, RewardModelArguments, TrainArgu
                 self.gradient_accumulation_steps = 1
             self.remove_unused_columns = False
             logger.info(f'Setting args.remove_unused_columns: {self.remove_unused_columns}')
-            self.truncation_strategy = 'left'  # Used for trimming the excessively long parts of a prompt.
+            if self.truncation_strategy != 'left':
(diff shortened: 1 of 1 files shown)


hjh0119:
Since the default value of truncation_strategy is raise, I think it would be better to add a logger.warning or logger.info here to alert the user. Alternatively, we could override the default value to left and then perform the check. I think both approaches would work.

What do you think?

Reichenbachian:
I mean, I think the ideal scenario here is we actually use the user specified truncation strategy, or we log it and don't allow the user to specify it at all. It seems like somewhat bad form to show the user an option which will be overwritten.

In our use case, raise/delete is much better. A left truncation strategy causes a crash for us (multimodal fine tuning with a cropped image token). Not sure why we overwrite here at all instead of just taking the cli arg. Any idea why that'd be?

hjh0119:
I think we can overload truncation_strategy in GRPOArgument and set a new default value of left, and raise an error when the user tries to modify it. Let me give it a try.

hjh0119:
should be fixed in https://github.com/modelscope/ms-swift/pull/3818

PR #3790, opened 2025-04-07
Fix model max length

Reichenbachian:
# PR type
- [x] Bug Fix

# PR information

Every time I train right now, regardless of what I set the `max_length`, `max_completion_length`, or `vllm_max_model_len` to, the max_model_len gets set to 8192. We have a hidden default in here that doesn't listen to the arguments that are set. I believe we want the engine to be set to `vllm_max_model_len`, but happy to change it if I'm wrong.

@hjh0119 @tastelikefeet


Reichenbachian:
I think my branch was out of date.

PR #4273, opened 2025-05-19
Support vit_lr aligner_lr

Jintao-Huang:


henserlu:
It seems like failure on Ascend 910B NPU platforms

PR #4295, opened 2025-05-21
fix n > 1 with vLLM V1 Engine

hjh0119:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Fix the issue where only one choice is returned when using V1 Engine and n > 1

https://github.com/vllm-project/vllm/pull/7381

The same logic can be found in vllm.LLM:
https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/entrypoints/llm.py#L1347-L1350

The response aggregation logic in the VLLM V1 Engine is detailed here:
https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/v1/engine/parallel_sampling.py#L101-L110

## Experiment results

### DIFF
--- a/swift/llm/infer/infer_engine/vllm_engine.py
+++ b/swift/llm/infer/infer_engine/vllm_engine.py
@@ -307,6 +307,12 @@ class VllmEngine(InferEngine):
         if kwargs.get('seed') is None:
             kwargs['seed'] = get_seed()
         res = SamplingParams(**kwargs)
+
+        if hasattr(res, 'output_kind'):
+            # fix n > 1 in V1 Engine
+            from vllm.sampling_params import RequestOutputKind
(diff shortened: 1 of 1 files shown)


hjh0119:
test code  (`VLLM_USE_V1 = 1`)
```python
    from swift.llm import InferEngine, InferRequest, RequestConfig
    from swift.plugin import InferStats
    model = 'Qwen/Qwen2.5-1.5B-Instruct'
    from swift.llm import VllmEngine
    engine = VllmEngine(model, gpu_memory_utilization=0.6, use_async_engine=False, max_model_len=8192)
    req = InferRequest(messages=[{'role': 'user', 'content': 'Hello!'}])
    output = engine.infer([req], RequestConfig(n=4))
```

Jintao-Huang (approved):

PR #4381, opened 2025-05-27
[megatron] fix save timeout

Jintao-Huang:


Jintao-Huang:
https://github.com/NVIDIA/Megatron-LM/issues/873

PR #4658, opened 2025-06-21
[gkd] support use_logits_to_keep/padding_free/packing & update gkd shell

Jintao-Huang:
fast.sh
100 its

llm 7B

截屏2025-06-21 22 22 07


Jintao-Huang:
vlm: 2B

padding_free: 3s/it
截屏2025-06-21 22 54 02

截屏2025-06-21 22 55 55

PR #4669, opened 2025-06-23
solve the default 'template_backend' bug in llm.tempalte.base.Templte._encode

hzg0601:
solve the template_backend bug: the correct default should be jinja rather than swift

# PR type
- [✅] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information
The default template_backend in llm.tempalte.base.Templte._encode should be jinja rather than swift, however, becuase the self.template_meta.template_type in a specific Template class are derived from MLLMTemplateType or LLMTemplateType, which is a specific type(e.g, ovis2) rather than 'dummy', the template_backend is always 'swift'.

## Experiment results

the prompt in vllm when inferring ovis: 
企业微信截图_17506653559250

the prompt in swift:
企业微信截图_17506654012609

### DIFF
--- a/swift/llm/template/base.py
+++ b/swift/llm/template/base.py
@@ -1114,7 +1114,7 @@ class Template(ProcessorMixin):
 
     def _encode(self, inputs: StdTemplateInputs) -> Dict[str, Any]:
         template_backend = self.template_backend
-        if (self.template_meta.template_type == 'dummy' and self.use_chat_template and not self.is_training
+        if (self.template_meta.template_type != 'dummy' and self.use_chat_template and not self.is_training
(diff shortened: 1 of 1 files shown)


hzg0601:
image
one can print these params to verify this PR

Jintao-Huang:
Hello 😊, what model is this? Why are there spaces between the tokens?

I feel like this is actually the core issue here.

By default, the template_backend should be "swift"; "jinja" needs to be set manually, since some models do not support the jinja template_backend.

Jintao-Huang:
`You are Qwen, created by Alibaba Cloud. You are a helpful assistant.`

This system message is not accurate.

PR #4806, opened 2025-07-02
Update bos_token

aacedar:
# PR type
- [] Bug Fix, [issue](https://github.com/modelscope/ms-swift/issues/4785)

Clipboard_Screenshot_1751457316

# PR information
1. qwen模型的encode 不管add_special_tokens=True/False，都不会输出special token,如 <|im_start|>, qwen模型这些特殊的token会在拼接的时候补上，无需额外处理
                所以需要判断bos_token是否为空
            2. 若bos_token不为空，则表示需要加specal token，bos_token = all_tokens[:idx]得到的是列表，在执行`prompts_text.append(''.join(res_context_list))`时 
                会因为res_context_list[0]的类型是list而报错，并且这里的bos_token是token_id而不是token，在encode时也会出错，所以采用self.tokenizer.bos_token更为合适

### DIFF
--- a/swift/llm/template/base.py
+++ b/swift/llm/template/base.py
@@ -1039,8 +1039,16 @@ class Template(ProcessorMixin):
             idx = all_tokens.index(single_token[0])
             bos_token = all_tokens[:idx]
             sep_token = all_tokens[idx + 1:]
+            """
+            1. qwen model tokenizer.encode function parameter add_special_tokens=True/False，both of then not output special token, forexample <|im_start|>, this special token will be added when process system/user/assistant message ,so should judge if bos_token is none/empty or not  
+            2. if bos_token is not none, means that current model should add specal token, old code has two errors:
+            	2.1 besides using bos_token = all_tokens[:idx],will get list as a element for res_context_list, but when execute code `prompts_text.append(''.join(res_context_list))`, will raise expcept
+            	2.2 element of res_context_list should be text（not token id）, bos_token = all_tokens[:idx] will get token_id list, this will error when execute tokenizer.encode()(encode token_id) 
+            	so we use self.tokenizer.bos_token is the most reasonable and correct
+            """
             if bos_token:
-                res_context_list.append(bos_token)
+                # res_context_list.append(bos_token)
+                res_context_list.append(self.tokenizer.bos_token)
                 res_context_types.append(ContextType.OTHER)
 
         if self.template_meta.is_post_system or not system:


hjh0119:
Nice catch!

Should we modify the function `_apply_chat_template_to_messages_list` to make it compatible with eos_token?

For example:

```diff
    def _apply_chat_template_to_messages_list(self, messages_list: InputsType):
        prompts_text = []
        for messages in messages_list:
            InferRequest.remove_response(messages)
            template_inputs, _ = StdTemplateInputs.from_dict({'messages': messages})
            res_context_list, _, _ = self.template._swift_encode(template_inputs)
-           prompts_text.append(''.join(res_context_list))
+           prompts_text.append(''.join(elem for elem in res_context_list if isinstance(elem, str)))

        return prompts_text
```

aacedar:
@hjh0119 I have read the full self.template._swift_encode(template_inputs) function, the return result is list, if the element of res_context_list is list, this is not corrent result, here should raise an exception instead of ignore the error
so, I think here not judge `if isinstance(elem, str)`, just throw an exception to remind user

aacedar:
if elem type is list, that means the elem is token_id value, not text value.
for example: tokenize.encode("<｜begin▁of▁sentence｜><｜end▁of▁sentence｜>" you will get token_id list, but it is str type befor tokenize.encode(). 
The elem of res_context_list should all be of type str, and then encoded uniformly in subsequent processing。

hjh0119:
I believe the related issue arises because `template_meta.prefix` contains the BOS token as an integer, which leads to a join failure. 

Therefore, I think ignoring integer values in the `_apply_chat_template_to_messages_list` method is a safer approach, since this method is only used for logging purposes and omitting the BOS token will not have any impact.

Are there any training scripts that fail to run on the main branch but work with this PR?

aacedar:
yes, the related issue arises because template_meta.prefix , I can not agree with you any more. you can see my other pr， this pr solve template_meta.prefix bug。https://github.com/modelscope/ms-swift/pull/4813

hjh0119:
so are there any training/infer scripts that fail to run on the main branch but work with this PR?

aacedar:
just replace your_own_path in --model, --output_dir, you can test


    CUDA_VISIBLE_DEVICES=7 \

    swift rollout \

      --model your_own_path/deepseek-ai/deepseek-coder-6.7b-base \
 
     > logs/log-grpo-rollout.log 2>&1 &



    CUDA_VISIBLE_DEVICES=0,1,2,3,4,5 \
 
    NPROC_PER_NODE=6 \
 
    swift rlhf \
 
      --rlhf_type grpo \

      --model your_own_path/deepseek-ai/deepseek-coder-6.7b-base \

      --reward_funcs accuracy \

      --use_vllm true \

      --vllm_mode server \

      --vllm_server_host 127.0.0.1 \

      --vllm_server_port 8000 \

      --train_type full \

      --torch_dtype bfloat16 \

      --dataset AI-MO/NuminaMath-TIR#1000 \

      --split_dataset_ratio 0 \

      --max_completion_length 512 \

      --num_train_epochs 1 \

      --per_device_train_batch_size 2 \

      --learning_rate 1e-6 \

      --gradient_accumulation_steps 2 \

      --save_total_limit 2 \

      --logging_steps 1 \

      --deepspeed zero2 \

      --max_length 4096 \

      --warmup_ratio 0.05 \

      --dataloader_num_workers 2 \

      --dataset_num_proc 4 \

      --num_generations 6 \

      --temperature 0.9 \

      --top_p 0.9 \

      --top_k 50 \

      --log_completions true \

      --num_iterations 1 \

      --beta 0.01 \

      --output_dir ./ds-base-grpo \

      --report_to tensorboard \

      > logs/log-grpo-test.log 2>&1 &

hjh0119:
I've tested the script you provided using your commit, but it still raises the following error: TypeError: sequence item 0: expected str instance, list found.

swift/llm/template/base.py
```diff
@@ -1039,8 +1039,16 @@ def _swift_encode(self, inputs: StdTemplateInputs):
             idx = all_tokens.index(single_token[0])
             bos_token = all_tokens[:idx]
             sep_token = all_tokens[idx + 1:]
+            """
```
tastelikefeet, line 1042: `auto_add_bos` is False for Qwen models
So the encode logic of Qwen models will not reach here

aacedar: this function will not work for Qwen models because the qwen.py:QwenTemplateMeta class set auto_add_bos = False, it will be raise exception for deepseek model

`swift/llm/template/template/qwen.py`
```
@dataclass
class QwenTemplateMeta(ChatmlTemplateMeta):
    default_system: Optional[str] = DEFAULT_SYSTEM
    auto_add_bos: bool = False
    stop_words: List[Word] = field(default_factory=lambda: ['<|endoftext|>'])
    agent_template: str = 'hermes'
```

aacedar:
> I've tested the script you provided using your commit, but it still raises the following error: TypeError: sequence item 0: expected str instance, list found.

you should fix #4806 bug first, I have commit pr for this bug

PR #4830, opened 2025-07-04
[Feature] SwanLab Lark callback

dykderrick:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates
- [ ] More Models or Datasets Support

# PR information
Similar to https://github.com/hiyouga/LLaMA-Factory/pull/7481, this PR integrates SwanLab Lark notification to `train-args`.

This PR also updates relevant web-ui and docs.

See Also: Issue https://github.com/modelscope/ms-swift/issues/4829.

## Experiment results

`cli`
image

`feishu`
image

`web-ui`
image

Issue #4829, opened 2025-07-04
SwanLab Notification Integration


dykderrick:
**Describe the feature**
SwanLab supports multiple notification callbacks post-training completion, such as DingTalk, Slack or Lark. I propose ms-swift supports such integrations so that developers can get immediate notifications when training completes.

**Paste any useful information**
Llama-Factory is already integrating some of them. See https://github.com/hiyouga/LLaMA-Factory/pull/7481.

dykderrick:
| 通知通道       | 文档链接 |  实现 |
|------------|----------|----------|
| Email      | [https://docs.swanlab.cn/plugin/notification-email.html](https://docs.swanlab.cn/plugin/notification-email.html) | |
| 飞书       | [https://docs.swanlab.cn/plugin/notification-lark.html](https://docs.swanlab.cn/plugin/notification-lark.html) | https://github.com/modelscope/ms-swift/pull/4830 |
| 钉钉       | [https://docs.swanlab.cn/plugin/notification-dingtalk.html](https://docs.swanlab.cn/plugin/notification-dingtalk.html) | |
| 企业微信   | [https://docs.swanlab.cn/plugin/notification-wxwork.html](https://docs.swanlab.cn/plugin/notification-wxwork.html) | |
| Discord    | [https://docs.swanlab.cn/plugin/notification-discord.html](https://docs.swanlab.cn/plugin/notification-discord.html) | |
| Slack      | [https://docs.swanlab.cn/plugin/notification-slack.html](https://docs.swanlab.cn/plugin/notification-slack.html) | https://github.com/modelscope/ms-swift/pull/4887 |

Jintao-Huang:
SwanLab is very easy to use, thank you!

Jintao-Huang:
https://github.com/modelscope/ms-swift/issues/4212

哈咯！这个features有时间帮忙支持嘛😊

PR #4887, opened 2025-07-09
[Feature] Add Swanlab Slack notification

dykderrick:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates

# PR information
Adds support for Swanlab notifications via Slack.

This change introduces the `swanlab_notification_methods` parameter, allowing users to specify which notification channels (Lark and/or Slack) should be used to receive experiment updates. It also includes corresponding parameters for the Lark and Slack webhook URLs and necessary logic to register the appropriate callbacks based on the selected notification methods.

Related PR https://github.com/modelscope/ms-swift/pull/4830 and Issue https://github.com/modelscope/ms-swift/issues/4829.

## Experiment results
`cli`
Screenshot 2025-07-09 at 16 32 21

`slack`
Slack 2025-07-09 16 48 58

`web-ui`
image

### DIFF
--- a/swift/llm/argument/train_args.py
+++ b/swift/llm/argument/train_args.py
@@ -1,7 +1,7 @@
 # Copyright (c) Alibaba, Inc. and its affiliates.
 import os
 from dataclasses import dataclass, field
-from typing import Literal, Optional
+from typing import List, Literal, Optional
 
 from transformers import Seq2SeqTrainingArguments
 from transformers.utils.versions import require_version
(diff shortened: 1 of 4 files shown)


swift/llm/argument/train_args.py
```diff
@@ -83,14 +89,20 @@ def _init_swanlab(self):
         if self.swanlab_token:
             swanlab.login(self.swanlab_token)
 
-        if self.swanlab_lark_webhook_url is not None:
+        if 'lark' in self.swanlab_notification_methods and self.swanlab_lark_webhook_url is not None:
```
Jintao-Huang, line 92: An error will occur if self.swanlab_notification_methods is None.

dykderrick:
```py
if self.swanlab_notification_methods and 'lark' in self.swanlab_notification_methods and self.swanlab_lark_webhook_url is not None:
```

Would this argument be too long? Do you think if `self.swanlab_notification_methods` variable is needed?

PR #5011, opened 2025-07-18
Fix NCCL timeout issues that may occur in multi-node training

cnyvfang:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Because the previous setting about ddp_timeout will not be executed in some cases. In the process of saving the eval or model checkpoint of multi-node multi-gpu training, there is a probability that the NCCL timeout will be caused by cluster performance and other problems, which will lead to the suspension of the training process. 

This PR solves this problem by synchronizing ddp_timeout parameters in different situations through environment variables.

## Experiment results


Jintao-Huang:
hello! 

use_torchacc will never return True, and this part of the code will be removed in the next version.

cnyvfang:
Thank you Jintao~ I will close this pr.

PR #5078, opened 2025-07-23
Fix: `model_type` may not read properly when it's not provided in training args

kiritoxkiriko:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Original model_type read model_type from CLI tool, traning config, arg.json and finally fallback into huggingface's model config

This step seem's reasonable, but it read `architectures` from models's config.json instead of `model_type`, nowadays most of models provide model_type natively, we don't have to infer model_type from `architectures`

For example, in Qwen3's config.json, we have `model_type`

Screenshot 2025-07-23 at 15 05 58

So I modify the _get_model_info, read `model_type` first, and then fallback into `architectures` if it's empty of not set.

## Experiment results

Minor change, no need

### DIFF
--- a/swift/llm/model/register.py
+++ b/swift/llm/model/register.py
@@ -456,9 +456,8 @@ def _get_model_info(model_dir: str, model_type: Optional[str], quantization_conf
         model_type = _read_args_json_model_type(model_dir)
     # check config first
     if model_type is None:
-        hf_model_type = HfConfigFactory.get_config_attr(config, 'model_type')
-        # incase it's empty str
-        model_type = hf_model_type if hf_model_type else None
+    if model_type is None:
+        model_type = HfConfigFactory.get_config_attr(config, 'model_type') or None
     if model_type is None:
         architectures = HfConfigFactory.get_config_attr(config, 'architectures')
         model_types = get_matched_model_types(architectures)


Jintao-Huang:
hello!
These two model_types have different meanings. The concept of model_type in ms-swift is different from that in transformers.

kiritoxkiriko:
> hello! These two model_types have different meanings. The concept of model_type in ms-swift is different from that in transformers.

Oh, so it may have different value with hf's `model_type` under certain condtion? Can you make an example fot this, since they share same value most of time.

My original goal is let swift fill `model_type` automaticlly if the value is not present. For example if I use qwen2.5-instruct it will return an error that let me pass `model_type` explicitly, i hope it could be derived from model config.json

kiritoxkiriko:
> hello! These two model_types have different meanings. The concept of model_type in ms-swift is different from that in transformers.

nerver mind, i see this in [Q&A](https://swift.readthedocs.io/zh-cn/latest/Instruction/%E6%94%AF%E6%8C%81%E7%9A%84%E6%A8%A1%E5%9E%8B%E5%92%8C%E6%95%B0%E6%8D%AE%E9%9B%86.html), thank you very much

kiritoxkiriko:
> nerver mind, i see this in [Q&A](https://swift.readthedocs.io/zh-cn/latest/Instruction/%E6%94%AF%E6%8C%81%E7%9A%84%E6%A8%A1%E5%9E%8B%E5%92%8C%E6%95%B0%E6%8D%AE%E9%9B%86.html), thank you very much

PR #5190, opened 2025-07-31
Support qlora merge-lora

Jintao-Huang:


wongzingji:
~Hello, not sure whether you can see the comment in a merged PR or not, but I found this feature is not included in the latest released version.
May I ask is this true, and what's the concern of not including it?
Would appreciate it if you could give some ideas!~
Sorry I realized this has been included into the current release already! I need to remove the current installed package manually or install with `-U`. thanks ^^

Jintao-Huang:
https://github.com/Jintao-Huang/llmscope/blob/main/examples/train/qlora/awq/merge_lora.sh

PR #5220, opened 2025-08-01
Fix: init_rope_scaling from model config first

mungg:
# PR type
- [ O ] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Updated the logic to parse `rope_scaling` from the `model config`, and to take `original_max_position_embeddings` to re-calculate `max_model_len`

## Experiment results

### DIFF
--- a/swift/llm/argument/base_args/model_args.py
+++ b/swift/llm/argument/base_args/model_args.py
@@ -109,10 +109,20 @@ class ModelArguments:
     def _init_rope_scaling(self):
         max_model_len = self.model_info.max_model_len
         try:
-            if isinstance(self.rope_scaling, str):
-                self.rope_scaling = json.loads(self.rope_scaling)
-                if 'factor' in self.rope_scaling and max_model_len:
+            if isinstance(self.model_info.rope_scaling, dict):
+                self.rope_scaling = self.model_info.rope_scaling 
+            elif isinstance(self.model_info.rope_scaling, str):
+                self.rope_scaling = json.loads(self.model_info.rope_scaling)
+            else:
+                raise ValueError(f'Model config has an invalid / empty rope_scaling')
+            if 'factor' in self.rope_scaling:
+                if 'original_max_position_embeddings' in self.rope_scaling:
+                    self.max_model_len = int(self.rope_scaling['original_max_position_embeddings'] * self.rope_scaling['factor'])
+                elif max_model_len:
                     self.max_model_len = int(self.rope_scaling['factor'] * max_model_len)
+                else:
+                    self.max_model_len = max_model_len
+            logger.info(f'max_model_len is set to {self.max_model_len} from the RoPE setting of model config')
         except Exception:
             assert self.max_model_len is not None, 'Use max_model_len together with rope_scaling'
             assert self.rope_scaling in ['linear', 'dynamic', 'yarn']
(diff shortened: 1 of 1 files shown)


swift/llm/argument/base_args/model_args.py
```diff
@@ -108,36 +108,49 @@ def _init_mixed_precision(self):
...
+        if 'factor' in rope_scaling:
+            rope_scaling_factor = rope_scaling['factor']
+        else:
+            rope_scaling_factor = max(float(math.ceil(self.max_model_len / max_model_len)), 1.0)
+        rope_scaling['factor'] = rope_scaling_factor
+        self.rope_scaling = rope_scaling
+        rope_model_len = int(max_model_len * rope_scaling_factor)
         logger.info(f'rope_scaling is set to type: {self.rope_scaling}')
-
+        if self.max_model_len:
```
tastelikefeet, line 140: Is `if self.max_model_len` nessarary? If no max_model_len given, the value should still override the `max_position_embeddings` in the config.json, right?

swift/llm/argument/base_args/model_args.py
```diff
@@ -108,36 +108,49 @@ def _init_mixed_precision(self):
...
+            self.max_model_len = rope_model_len
+            assert self.max_model_len <= rope_model_len, f'rope config ({rope_model_len}) should be bigger than max_model_len from command line ({self.max_model_len})'
```
tastelikefeet: `self.max_model_len = rope_model_len` will always make `assert self.max_model_len <= rope_model_len` be true

swift/llm/argument/base_args/model_args.py
mungg: You are right! I fixed them in the new commit :D 

I thought of two possible cases:

When the user provides max_model_len via the command line (`if self.max_model_len`)

In this case, they likely also provided the rope (yarn) setting via the command line. If the assertion fails, it's probably because the factor in the yarn setting is incorrect. This can happen if the factor is too small to support the desired max_model_len.

alternatively, the user may not have set yarn via the command line, but the model_config includes a yarn setting. I assume this isn't the scenario you intended since you recommend explicitly combining yarn with max_model_len. Still, in this case, I added a check to notify the user if the expected rope-extended length from the model config is smaller than the desired max_model_len.

When the user doesn’t provide max_model_len: (`else`)
In that case, we probably rely entirely on what's defined in the model config.

swift/llm/argument/base_args/model_args.py
mungg: thank you! I removed it and add else statement in new commit ;p

swift/llm/argument/base_args/model_args.py
tastelikefeet: I think there is one case missed:
When the user offered a dict type `rope_scaling` in the cli like `{type:yarn,factor:2.0}`, in this case the max_model_len may not be set(because the max_model_len can be set via the factor), then the condition will cause an error, isn't it?

mungg: I see. I moved the max_model_len check (`assert self.max_model_len is not None, 'Use max_model_len together with rope_scaling'`) to only apply when rope_scaling is given as just a type string!

tastelikefeet (approved):

PR #5305, opened 2025-08-08
[model] support glm-4.5 agent

laocheyujie:
# PR type
- [ ] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

This PR adds support for GLM-4.5 agent (function call) training in ms-swift.

Key updates:
1. Added GLM4_5AgentTemplate in plugin/agent_template/glm4.py to support training GLM-4.5 function call agents.
2. Enhanced suffix appending logic: in the final generation round, check if the output already ends with a stop word, and skip appending the suffix if so to avoid redundancy.

## Experiment results
- Training successfully runs on 8×A100 (80GB) with GLM-4.5 agent configuration.

### DIFF
--- a/swift/llm/template/template/glm.py
+++ b/swift/llm/template/template/glm.py
@@ -65,6 +65,10 @@ class GLM4_0414TemplateMeta(GLM4TemplateMeta):
     agent_template: str = 'glm4_0414'
 
 
+@dataclass
+class GLM4_5TemplateMeta(GLM4_0414TemplateMeta):
+    agent_template: str = 'glm4_5'
+
 class GLM4_1VTemplateMeta(GLM4_0414TemplateMeta):
     system_prefix: Optional[Prompt] = field(default_factory=lambda: ['[gMASK]<sop><|system|>{{SYSTEM}}'])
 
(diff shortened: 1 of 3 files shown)


swift/llm/template/base.py
```diff
@@ -1135,7 +1135,8 @@ def _swift_encode(self, inputs: StdTemplateInputs):
                 context_list.append('{{RESPONSE}}')
                 # self.is_training needed because we may want to continue generation from
                 # the current response
-                if self.is_training and not sep_token or self.task_type == 'embedding':
+                string_stop_words = tuple(s for s in template_meta.stop_words if isinstance(s, str))
```
Jintao-Huang: What is the reason for this modification here?

laocheyujie: Originally, I made that change to prevent models like GLM-4.5 from producing `<|observation|><|user|>` when the final round was a tool_call. But I just realized this can be handled more cleanly by overriding `_swift_encode` in the `GLM4_5Template` class, so I removed that part of the modification in my latest commit.

Jintao-Huang:
Hello, please run the following code to pass the lint test.
```
pip install pre-commit
pre-commit run --all-files
```

swift/llm/template/base.py
Jintao-Huang: Hello!
Here, `<|user|>` (suffix) is only added during training. Could you please clarify: in what training scenarios would the final round include `<|observation|>` but not tool_response?

PR #5426, opened 2025-08-18
[model] Support ovis2.5

Jintao-Huang:
https://github.com/AIDC-AI/Ovis/issues/58


MooMoo-Yang:
So, can we infer && finetune ovis-2.5 with ms-swift now?

PR #5518, opened 2025-08-25
[template] update glm4.5 agent template

Jintao-Huang:


Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/5305

PR #5519, opened 2025-08-25
[bugfix] fix rstrip

Jintao-Huang:


Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/5518

Issue #5575, opened 2025-08-28
deepeyes 报错 dacite.exceptions.UnionMatchError: can not match type "NoneType" to any type of "messages" union: typing.Union[str, typing.List[typing.Dict[str, typing.Any]], typing.List[int]]


guihonghao:
https://github.com/modelscope/ms-swift/blob/main/docs/source/Instruction/GRPO/AdvancedResearch/deepeyes.md
按照官方文档下载数据集，部署，启动训练模型，报错，怎么解决？谢谢。
[rank0]: dacite.exceptions.UnionMatchError: can not match type "NoneType" to any type of "messages" union: typing.Union[str, typing.List[typing.Dict[str, typing.Any]], typing.List[int]]
Train:   0%|                                                                      | 0/47052 [00:04<?, ?it/s]

```
CUDA_VISIBLE_DEVICES=0,1,2,3 \
swift deploy \
    --model /data/guihonghao/models/Qwen2.5-VL-72B-Instruct \
    --vllm_tensor_parallel_size 4
```

```
MAX_PIXELS=602112 \
CUDA_VISIBLE_DEVICES=4 \
swift rollout \
    --model /data/guihonghao/models/Qwen2.5-VL-7B-Instruct \
    --vllm_use_async_engine true \
    --external_plugins examples/train/grpo/plugin/deepeyes/deepeyes_plugin.py  \
    --multi_turn_scheduler deepeyes_scheduler \
    --vllm_max_model_len 8192 \
    --vllm_gpu_memory_utilization 0.9 \
    --max_turns 5
```

```
MAX_PIXELS=602112 \
CUDA_VISIBLE_DEVICES=5,6,7 \
NPROC_PER_NODE=3 \
swift rlhf \
    --rlhf_type grpo \
    --model /data/guihonghao/models/Qwen2.5-VL-7B-Instruct \
    --dataset "/data/guihonghao/data/DeepEyes-Datasets-47k/data_0.1.2_visual_toolbox_v2.parquet" \
    "/data/guihonghao/data/DeepEyes-Datasets-47k/data_thinklite_reasoning_acc.parquet" \
    "/data/guihonghao/data/DeepEyes-Datasets-47k/data_v0.8_visual_toolbox_v2.parquet" \
    --use_vllm true \
    --vllm_mode server \
    --vllm_server_host 127.0.0.1 \
    --vllm_server_port 8001 \
    --offload_optimizer true \
    --offload_model true \
    --sleep_level 1 \
[...]

PR #5576, opened 2025-08-28
[template] fix suffix

Jintao-Huang:


Jintao-Huang:
https://github.com/modelscope/ms-swift/issues/5575

PR #5645, opened 2025-09-03
bug fix: RuntimeError when training GRPO with LoRA and PtEngine

chenjianhuii:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information
During PEFT training, to obtain the logps of the reference model, `GRPOTrainer` uses `null_ref_context()` to temporarily disable PEFT. Upon exiting this context, re-enabling PEFT requires setting `requires_grad_(True)` on each layer. However, the model was previously used in `PtEngine` within the context of `@torch.inference_mode()`, which results in the error **"RuntimeError: Setting requires_grad=True on inference tensor outside InferenceMode is not allowed."** Switching from `@torch.inference_mode()` to `@torch.no_grad()` can resolve this issue but may lead to a trade-off in performance. I'm uncertain if there is a better solution.

### DIFF
--- a/swift/llm/infer/infer_engine/pt_engine.py
+++ b/swift/llm/infer/infer_engine/pt_engine.py
@@ -461,7 +461,7 @@ class PtEngine(InferEngine):
             return await queue.get()
 
     # Ensure `template._post_encode` has no gradient.
-    @torch.no_grad()
+    @torch.inference_mode()
     def _infer(
         self,
         infer_requests: List[InferRequest],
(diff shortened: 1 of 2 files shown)


hjh0119:
can't produce with the example script https://github.com/modelscope/ms-swift/blob/main/examples/train/grpo/internal/pt.sh

Could you open an issue that includes this reproduction script along with your environment details?

hjh0119:
I opted for a more subtle modification, switching from the no_grad context to inference_mode for get_logps. 

Can you retry?

chenjianhuii:
> I opted for a more subtle modification, switching from the no_grad context to inference_mode for get_logps.
> 
> Can you retry?

It raise new error
image

PR #5696, opened 2025-09-07
[Megatron] Convert Megatron adapter to Huggingface adapter without merging

casper-hansen:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

It is currently not possible to convert a Megatron LoRA adapter to Huggingface adapter without merging into the full model. Having only the adapter weights is useful in the case of multi-lora inference.

TODO:
- generalize LoraConfig

## Experiment results

TBD
- running perplexity measurements against private test set


casper-hansen:
Closing this as I never could get it working. Would like someone to implement this, but I don't have the time to carry it through.

PR #5773, opened 2025-09-11
[grpo] Optimize vLLM weight synchronization & update buitin accuracy reward

hjh0119:
### LoRA
1. Synchronize/load only the trained adapter weights.(both colocate / server mode), for server mode, use `--vllm_enable_lora true` in rollout
2. In server mode, transmit flattened adapter weights to reduce communication overhead of model parameters.

### FULL
1. Remove the original per-tensor synchronization logic and adopt a bucketing strategy to reduce redundant communication requests and overhead, especially for **MoE** models (which have more tensors than dense models).‘
2. Removed per-tensor gather and reverted to using batched gather; as a result, `move_model_batches` now works with full-parameter training.

## Update: Built-in Accuracy Reward

- Fixed several cases where accuracy could not be correctly evaluated
- Added corresponding unit tests.

### DIFF
--- a/docs/source_en/Instruction/GRPO/GetStarted/GRPO.md
+++ b/docs/source_en/Instruction/GRPO/GetStarted/GRPO.md
@@ -202,7 +202,7 @@ Swift 3.10 optimizes weight synchronization, and setting the following parameter
 # rollout(server mode)
 swift rollout \
     --vllm_enable_lora true \
-    --vllm_max_lora_rank xxx # 与训练脚本lora_rank一致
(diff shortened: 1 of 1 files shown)


hjh0119:
Qwen2.5-VL-7B-Instruct, server mode, tp=2, dp=2 → 10× speed-up

image

hjh0119:
Qwen2.5-VL-32B-Instruct(**LoRA Training**) Reduced from **4s** to **1s**.

> lora+ indicates that only the LoRA adapter weights are synchronized, and the parameter --vllm_enable_lora true is used on the rollout side.

image

PR #5777, opened 2025-09-11
feat: Add DeepSeek V3.1 Agent Template Support

gakkiri:
- Add DeepSeekV31AgentTemplate class with DSV3.1 specific tool call format
- Support <｜tool▁calls▁begin｜><｜tool▁call▁begin｜>name<｜tool▁sep｜>args<｜tool▁call▁end｜> format
- Add agent_template='dsv3_1' parameter to deepseek_v3_1 template registration
- Update agent_templates registry to include 'dsv3_1' template
- Add comprehensive test cases for dsv3_1 agent template functionality
- Maintain backward compatibility with react_en format as fallback

# PR type
- [ ] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

his PR adds support for DeepSeek V3.1 Agent Template with the new tool calling format.

## Experiment results

- [x] Added unit tests for dsv3_1 agent template
- [x] Verified tool call parsing with DSV3.1 format

### DIFF
--- a/tests/test_align/test_template/test_agent.py
+++ b/tests/test_align/test_template/test_agent.py
@@ -380,7 +380,7 @@ def test_qwen3_coder():
 def test_dsv3_1():
     agent_template = agent_templates['dsv3_1']()
 
-    engine = PtEngine('/apdcephfs_sh9/share_303378293/hunyuan/public/model/DeepSeek-V3.1-bf16', device_map='cpu', model_type='deepseek_v3_1')
+    engine = PtEngine('deepseek-ai/DeepSeek-V3.1', device_map='cpu', model_type='deepseek_v3_1')
     template = engine.default_template
     template.agent_template = agent_template
 
--- a/swift/plugin/agent_template/dsv3_1.py
+++ b/swift/plugin/agent_template/dsv3_1.py
@@ -16,7 +16,7 @@ class DeepSeekV31AgentTemplate(BaseAgentTemplate):
     def get_toolcall(self, response: str) -> List['Function']:
         from swift.llm.infer import Function
         # Parse tool calls using the DSV3.1 format: <｜tool▁calls▁begin｜><｜tool▁call▁begin｜>name<｜tool▁sep｜>args<｜tool▁call▁end｜>
-        pattern = r'<｜tool▁call▁begin｜>([^<]+)<｜tool▁sep｜>([^<]+)<｜tool▁call▁end｜>'
+        pattern = r'<｜tool▁call▁begin｜>(.*?)<｜tool▁sep｜>(.*?)<｜tool▁call▁end｜>'
         res_list = re.findall(pattern, response, re.DOTALL)
         functions = []
         for name, arguments in res_list:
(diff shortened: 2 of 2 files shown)


Jintao-Huang:
hello!

please run the following commands: 😊
```shell
pip install pre-commit
pre-commit run --all-files
```

tests/test_align/test_template/test_agent.py
```diff
@@ -344,15 +344,73 @@ def test_glm4_5():
     template.template_backend = 'swift'
     template.set_mode('train')
     encoded = template.encode(data)
-    print(f'input_ids: {template.safe_decode(encoded["input_ids"])}')
-    print(f'labels: {template.safe_decode(encoded["labels"])}')
+    print(f'input_ids: {template.safe_decode(encoded['input_ids'])}')
```
Jintao-Huang: Hi, could you revert the single/double-quote changes here? I can help you get it through the lint tests — there's a syntax error here.

gakkiri: So after running pre-commit, it went and swapped all the "" in my f-strings to '', which totally broke stuff. Ended up rolling it back.

Since pre-commit only fixes f-strings in this PR, I think it is no longer necessary.

swift/llm/template/template/deepseek.py
```diff
@@ -245,18 +245,49 @@ class DeepseekV2_5TemplateMeta(TemplateMeta):
...
+
 
 register_template(
     DeepseekV2_5TemplateMeta(LLMTemplateType.deepseek_r1, template_cls=ThinkingTemplate, response_prefix='<think>\n'))
 
 # enable thinking: response_prefix='<think>'
 register_template(
     DeepseekV2_5TemplateMeta(
-        LLMTemplateType.deepseek_v3_1, template_cls=DeepseekV3_1Template, response_prefix='</think>'))
+        LLMTemplateType.deepseek_v3_1, template_cls=DeepseekV3_1Template, response_prefix='', agent_template='dsv3_1'))
```
Jintao-Huang, line 290: What’s the reason for removing response_prefix here?😊

swift/llm/template/template/deepseek.py
```diff
@@ -245,18 +245,49 @@ class DeepseekV2_5TemplateMeta(TemplateMeta):
 register_template(DeepseekV2_5TemplateMeta(LLMTemplateType.deepseek_v2_5))
 
 
-class DeepseekV3_1Template(ThinkingTemplate):
+class DeepseekV3_1Template(Template):
```
Jintao-Huang, line 248: What’s the reason for switching from ThinkingTemplate to Template here?

gakkiri: Because the agent template of v3.1 is special, for example:
```
<｜begin▁of▁sentence｜><｜User｜>北京天气<｜Assistant｜></think>I'll get the current temperature for Beijing for you.<｜tool▁calls▁begin｜><｜tool▁call▁begin｜>get_current_temperature<｜tool▁sep｜>{"location": "Beijing, China"}<｜tool▁call▁end｜><｜tool▁calls▁end｜><｜end▁of▁sentence｜><｜tool▁output▁begin｜>29摄氏度, 晴<｜tool▁output▁end｜>北京现在的天气是：
- 温度：29摄氏度
- 天气状况：晴天

天气很不错，适合外出活动！<｜end▁of▁sentence｜><｜User｜>ok, good<｜Assistant｜></think>Great! I'm glad I could help you check the weather in Beijing. It does sound like a nice day with 29°C and sunny skies. If you need to check the weather for any other locations or dates, just let me know!<｜end▁of▁sentence｜>
```

You will find that the `content` after `tool_response (<｜tool▁output▁end｜>)` cannot contain `Assistant` and `</think>`, so the logic cannot be universal

Here, copy ThinkingTemplate and add if judgment for this point in
```
                    if not message['content'].startswith(('<think>', self.no_think_prefix)):
                        if i > 0 and messages[i - 1]['role'] not in ['tool', 'tool_call']:
                            message['content'] = self.no_think_prefix + message['content']
```

swift/llm/template/template/deepseek.py
gakkiri: Due to the reasons mentioned in the second comment, I had to remove the `</think>` in the response when facing tool call.

This may affect the `</think>` in normal conversations. I think that since this agent template is too special, maybe I should write a new class to handle it.

PR #6182, opened 2025-10-17
support mcore_bridge

Jintao-Huang:
Features:
1. During training, weights can be loaded directly from safetensors format and saved as safetensors weights. This eliminates the model conversion step.
2. Multi-node weight conversion support, enabling adaptation for extremely large models.
3. GRPO support: streaming generation of safetensors format state_dict from megatron model.
4. LoRA weight export


arvyanh:
Is there a plan for when this will be implemented? Facing some disk space and efficiency issue doing the 235B and 671B models lately
I could try to implement it if not doing that right now?

PR #6325, opened 2025-10-28
[algo] support RLOO algorithm 

hjh0119:
### 1. Support [RLOO](https://arxiv.org/pdf/2402.14740) in GRPOTrainer with the following new arguments:

- advantage_estimator: rloo – Enables the RLOO advantage estimation strategy.
- kl_in_reward: true – Includes KL penalty in the reward computation, as required by RLOO.

### 2. Enhanced Advantage Normalization with Batch-Size Support

Refactored the `scale_rewards` parameter to support multiple normalization modes. It now accepts three options: [group, batch, none], replacing the previous boolean flag:

- group: Normalizes advantages using the standard deviation within each group (equivalent to the old True behavior).
- batch: New option that normalizes advantages using the standard deviation across the entire batch.
- none: Disables normalization (equivalent to the old False behavior).

https://github.com/modelscope/ms-swift/issues/4919
https://github.com/modelscope/ms-swift/issues/5721


hjh0119:
A similar experimental performance comparison with GRPO.

image

PR #6356, opened 2025-10-29
[bugfix] fix megatron qwen3_vl gradient_checkpointing

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/6352


mertunsall:
Hi @Jintao-Huang - what would be the best way of having this change in the docker image? it is very hard to build the environment

PR #6461, opened 2025-11-06
[docs] update docs

Jintao-Huang:


mertunsall:
Hi @Jintao-Huang 

I was using the previous image you pushed and fine-tuned the 8B dense qwen3-vl as well as 30b-a3b models with megatron, same data and 3 epochs on same hardware. But when I eval performance of the 30B MoE is actually worse than that of 8B. What could it be?

PR #6504, opened 2025-11-09
Add Tensor Input Support: Enable .pt file processing with <tensor> tags for latent representations 

Marshall-mk:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

## Experiment results


Jintao-Huang:
Hello. Thanks for your PR. Could you tell me what additional advantages using `<tensor>` has compared to using `<image>` ?

Marshall-mk:
> Hello. Thanks for your PR. Could you tell me what additional advantages using `<tensor>` has compared to using `<image>` ?

The key difference here is that you can't load tensors (.pt files) using the <image> or <video> tags, since the file extensions for these types are not supported by the respective formats. The supported video extensions are: {.mp4, .mkv, .mov, .avi, .wmv, .flv, .webm} and for images: {.jpg, .jpeg, .png, .gif, .bmp, .tiff, .webp}.

You might be wondering why we’d want to load tensors at all. In my case, I work with medical images such as CT or MRI scans, which often contain multiple slices and have large dimensions. These images make training VLMs very compute-intensive. To efficiently process these images, we need to compress them using an encoder, save the resulting embeddings as tensors, and then use the VLMs available in ms-swift.

Now, a naive approach would be to convert the embeddings (the tensors) into videos (from .pt files to .mp4), but that introduces the need for additional normalizations and scaling, which actually leads to some disparity in the actual embeddings. This is why using <tensor> directly is a better approach, as it avoids the unnecessary conversion and preserves the embeddings.

PR #6952, opened 2025-12-08
[Chord] bug fix in phi loss

Birdylx:
# PR type
Bug Fix

# PR information
fix phi loss bug, the coeff should be stoped gradient as paper describe

## Experiment results


hjh0119:
Thank you for your contribution

Could you share a reference for why the phi function should stop gradients?

Birdylx:
@hjh0119 here is the original implementation https://github.com/modelscope/Trinity-RFT/blob/613194d45fee0eef9145fb73dbda69cab17fd6f4/trinity/algorithm/policy_loss_fn/chord_policy_loss.py#L97
phi is the weight based on logprob, we do not want gradient for this, it is not learnable.

PR #7019, opened 2025-12-12
[bugfix] fix missing generate method for InternVL-2.5

xwy-bit:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

**Summary** This PR addresses a compatibility issue with `transformers` version **4.50+**, where `PreTrainedModel` no longer inherits from `GenerationMixin` by default. As a result, the InternLM2 model loses its text generation capability, causing an `AttributeError`.

Changes

- Modified `get_model_tokenizer_internvl` to perform a runtime check on the loaded model.
- Dynamically injects `GenerationMixin` into the model's `__bases__` if the `generate` method is missing.
- Ensures backward compatibility with older `transformers` versions.

## Before vs. After
**Before**

image

**After**
image

### DIFF
--- a/swift/llm/model/model/internlm.py
+++ b/swift/llm/model/model/internlm.py
@@ -5,7 +5,7 @@ from typing import Any, Dict
 
 import torch
 from transformers.dynamic_module_utils import get_class_from_dynamic_module
-
+from transformers import GenerationMixin
 from swift.llm import TemplateType
 from ..constant import LLMModelType, MLLMModelType, RMModelType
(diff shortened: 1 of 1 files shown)


Jintao-Huang:
try use transformers==4.49

xwy-bit:
You're right. I hope this PR remains compatible with both versions prior to 4.49 and above, since 4.50 introduced the breaking change, and transformers is already at 4.57.3 (5.0.0rc).

PR #7078, opened 2025-12-16
[npu] Fix the failure in mcore version check on NPU device

tongtong0613:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

When using Mindspeed + Megatron on NPU devices, `version.parse(importlib.metadata.version('megatron-core'))` fails to correctly retrieve the mcore version and raises the error:

`importlib.metadata.PackageNotFoundError: No package metadata was found for megatron-core`.

So, skip `_patch_unified_memory` on NPU devices temporarily.

## Experiment results

### DIFF
--- a/swift/megatron/init.py
+++ b/swift/megatron/init.py
@@ -17,6 +17,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 from packaging import version
 from tqdm import tqdm
+from transformers.utils import is_torch_npu_available
 
 from swift.llm import git_clone_github
 from swift.utils import (get_logger, is_flash_attn_3_available, is_megatron_available, safe_ddp_context, split_list,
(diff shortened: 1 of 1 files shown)


swift/megatron/init.py
```diff
@@ -791,7 +791,9 @@ def _apply_rotary_pos_emb_thd(
 
 
 def _patch_unified_memory():
-    mcore_015 = version.parse(importlib.metadata.version('megatron-core')) >= version.parse('0.15.0rc0')
+    import megatron.core
```
Jintao-Huang: This will cause megatron.core to be loaded, which makes this patch ineffective.

How about just skipping the patch for NPU directly?

tongtong0613: Ok, I'll update this PR.

Issue #7132, opened 2025-12-19
vLLM 0.13.0 causes error due to deprecated VLLM_ATTENTION_BACKEND environment variable


3manifold:
# Cmdline 

```bash
  VLLM_USE_V1=0 \
  FORCE_QWENVL_VIDEO_READER=torchcodec  \
  USE_AUDIO_IN_VIDEO=True \
  ENABLE_AUDIO_OUTPUT=False \
  FPS_MAX_FRAMES=4 \
  swift infer \
      --model "Qwen3-Omni-30B-A3B-Instruct" 
      --infer_backend vllm \
      --val_dataset "$WORK_DIR/query.jsonl" \
      --result_path "$WORK_DIR/captions_Qwen3-Omni-30B-A3B-Instruct.jsonl"

```

# Error
```
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/vllm/vllm_flash_attn/flash_attn_interface.py", line 253, in flash_attn_varlen_func
[rank0]:     out, softmax_lse = torch.ops._vllm_fa2_C.varlen_fwd(
[rank0]:                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/opt/conda/lib/python3.11/site-packages/torch/_ops.py", line 1255, in __call__
[rank0]:     return self._op(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: torch.AcceleratorError: CUDA error: the provided PTX was compiled with an unsupported toolchain.
[rank0]: Search for `cudaErrorUnsupportedPtxVersion' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
[rank0]: CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
[rank0]: For debugging consider passing CUDA_LAUNCH_BLOCKING=1
[rank0]: Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

[rank0]:[W1219 15:23:53.162823414 ProcessGroupNCCL.cpp:1524] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
[...]

PR #7133, opened 2025-12-19
Improve vLLM examples regarding vllm_engine_kwargs use 

3manifold:
# PR type 
- [x] Document Updates 

# PR information
Add an example of an advanced usecase of `--vllm_engine_kwargs ` use since there was no similar example in the past.  It applies to `vllm >= 0.13.0`.
List of input vLLM args can be found here https://github.com/vllm-project/vllm/pull/26315 . It relates to https://github.com/modelscope/ms-swift/issues/7132 .


3manifold:
> vllm         | >=0.5.1      | 0.11.0/0.13.0       | Inference/Deployment

Given that the current [README.md](https://github.com/modelscope/ms-swift/blob/main/README.md)  suggests `vllm = 0.13.0`,  the example in this PR can be quite helpful.

Issue #7236, opened 2025-12-29
support QwenLong-L1.5-30B-A3B


pruidong:
**Describe the feature**

support [QwenLong-L1.5-30B-A3B](https://www.modelscope.cn/models/iic/QwenLong-L1.5-30B-A3B)

jpg1024:
QwenLong-L1.5- 30B-A3B Can I use qwen2 's dialogue template for reasoning?

thanks.

PR #7237, opened 2025-12-29
[model] support qwenlong L1.5

Jintao-Huang:


Jintao-Huang:
https://github.com/modelscope/ms-swift/issues/7236

PR #7278, opened 2026-01-04
feat(cli): add setproctitle support to customize process name

ciaoyizhen:
This commit introduces the setproctitle library to allow users to customize the system process name. Users can set the process title via --proc_title argument or environment variable.

# PR type
- [x] New Feature

# PR information

This PR introduces the `setproctitle` library to allow users to customize the system process name for training and inference tasks. This is particularly useful for distinguishing between multiple running experiments on the same machine when using tools like `top` or `ps`.

### Key Changes:
1.  **Dependency**: Added `setproctitle` to `requirements/framework.txt`.
2.  **Implementation**: Added `try_get_proc_title` in `swift/cli/utils.py`. The function handles the dependency gracefully (it uses `try-except` to avoid crashes if the library is missing).
3.  **Logic**: The process title is determined with the following priority:
    * **CLI Argument**: `--proc_title <name>`

    * **Default**: `'ms-swift'`
4.  **Entry Points**: Applied this logic to `sft.py`, `pt.py`, and `rlhf.py`.

### DIFF
--- a/swift/llm/argument/base_args/base_args.py
+++ b/swift/llm/argument/base_args/base_args.py
@@ -61,6 +61,7 @@ class BaseArguments(CompatArguments, GenerationArguments, QuantizeArguments, Dat
 
     Args:
         tuner_backend (str): The tuner backend to use. Choices are 'peft' or 'unsloth'. Default is 'peft'.
+        proc_title (str): The proc title name to be set. Default is 'ms-swift'.
         train_type (str): The training type. Choices include 'lora', 'full', 'longlora', 'adalora', 'llamapro',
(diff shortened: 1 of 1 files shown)


ciaoyizhen:
上面偷懒 pr描述让gemini生成 然后修改了一下

总的来说是这样的，有些公司会要求你启动一个训练，以及启动一个服务(占显存)，需要使用进程名来控制，这样方便定位。
然后我想，这是一个端到端的训练，所以我觉得需要补一个训练的进程名。
然后这个代码我试过sft的，没问题，然后其他的如rlhf和pt，理论上是一样的，就没测过

PR #7279, opened 2026-01-04
Fix the issue of repeated inference in multi-turn scheduler.

Simon-ss7:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# Bug information

version: 3.12.0
infer_engine: vllm
model: qwen3

When I tried to follow the demo under example/train/grpo/plugin.py/ToolCallScheduler to run multi-turn training, I encountered the following issues:

In colocate mode, the launch fails with a null pointer error indicating that self.infer_engine is None.

In server mode, when current_turn > 1, the returned responses contain duplicated inference results. After inspecting the _add_request method in vllm_engine.py, I found that the inference content for each step does not append the results from previous steps.

# PR information

I found that in colocate mode, a tokenizer is passed when initializing the multi_turn_scheduler in rollout_mixin.py, and in server mode, self.tokenizer is also available inside multi_turn_scheduler. Based on this, I adjusted the logic for obtaining the tokenizer in the step method of plugin.py/ToolCallScheduler.

I found that in server mode, the run method of MultiTurnScheduler defines {'role': 'assistant', 'content': None} at lines 352–354 to append inference results, but when current_turn > 1, this entry is removed at lines 236–238. As a result, when I inspected the _add_request method in vllm_engine.py, the input_ids always corresponded to the initial inference content.

After adjusting the logic in the run method so that the appended assistant=None entry is not removed when current_turn > 1, the inference content became correct.

## Experiment results

colocate_step_error
server_repeat_error

### DIFF
--- a/swift/plugin/multi_turn.py
+++ b/swift/plugin/multi_turn.py
@@ -235,6 +235,7 @@ class MultiTurnScheduler(RolloutScheduler, ABC):
             messages = current_request.messages
             if current_turn == 1:
                 # If it's the first turn, remove the response
+                # Keep the original logic, but I think this step is unnecessary here.
                 remove_response(messages)
 
             # Get model response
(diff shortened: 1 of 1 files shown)


hjh0119:
LGTM
thanks

PR #7299, opened 2026-01-06
fix npu megatron cp

addsubmuldiv:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information
## Summary

Fix a MindSpeed context-parallel (ring-attn) crash on **NPU** when running **swift-megatron**.
Root cause is that swift-megatron passed an incomplete `PackedSeqParams` object into MindSpeed: two padded cu-seqlens fields were not initialized, which made MindSpeed take a wrong branch in `ring_context_parallel.forward` and crash (stack trace as shown in the screenshot).

This PR initializes the missing fields **only on NPU** to align with MindSpeed’s expected parameter contract.

## Background / Issue

When enabling MindSpeed context parallel / ring attention on Ascend NPU, training may crash in the following call chain:

image

The error is caused by MindSpeed reading `PackedSeqParams` fields that are expected to exist in CP code path.

## Root Cause

In swift-megatron, the `PackedSeqParams` object constructed from `position_ids` only had basic fields filled (e.g., `cu_seqlens_q`, `cu_seqlens_kv`, `max_seqlen_*`, `qkv_format`).
However, for NPU + MindSpeed CP path, MindSpeed also expects the padded versions:

* `cu_seqlens_q_padded`
* `cu_seqlens_kv_padded`

When these fields are missing/`None`, MindSpeed’s internal branching logic is misled and enters an incompatible path, leading to a runtime crash in `ring_context_parallel.forward`.

## Fix

* Construct `PackedSeqParams` and **explicitly set**:

  * `packed.cu_seqlens_q_padded`
  * `packed.cu_seqlens_kv_padded`
* The assignment is guarded by `is_torch_npu_available()` so non-NPU behavior is unchanged.

### 1) Functional

* NPU environment no longer crashes when running swift-megatron + MindSpeed context parallel/ring-attn.

### 2) Numerical sanity check (GPU vs NPU)

We compared the first **100 global steps** loss curves between **GPU** and **NPU (with this fix)**.

* The two curves show consistent convergence trend.
* Mean Absolute Error (MAE) across the first 100 steps: **~0.00665** (N=100).
* Small per-step differences are expected due to bf16 kernels / non-determinism across devices, but overall behavior matches well.

> Loss curve figure: attached in PR description (GPU vs NPU, first 100 steps).

## How to reproduce (before fix)

1. Run on Ascend NPU with swift-megatron.
2. Enable tensor parallel、 context parallel / ring attention.
3. Training crashes inside `mindspeed/core/context_parallel/ring_context_parallel.py` due to wrong branch triggered by incomplete `PackedSeqParams`.

## Experiment results

image

### DIFF
--- a/swift/megatron/trainers/utils.py
+++ b/swift/megatron/trainers/utils.py
@@ -15,6 +15,7 @@ from megatron.core.optimizer import ChainedOptimizer
 from megatron.core.packed_seq_params import PackedSeqParams
 from megatron.training import get_args, get_wandb_writer
 from packaging import version
+from transformers.utils import is_torch_npu_available
 
 from swift.llm import get_packed_seq_params as _get_packed_seq_params
 from swift.llm import to_device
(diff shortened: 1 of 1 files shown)


Jintao-Huang (approved):

Jintao-Huang:
thanks

PR #7348, opened 2026-01-09
feat: support scale rewards "gdpo"

Auraithm:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

Add GDPO (Group reward-Decoupled Normalization Policy Optimization) trainer support for Megatron.

## GDPO vs GRPO

**GRPO**: Sum rewards first, then normalize

adv = normalize(r1 + r2)

**GDPO**: Normalize each reward separately, then sum, then batch-level normalize

adv = batch_normalize(normalize(r1) + normalize(r2))

GDPO preserves fine-grained differences between reward combinations. For example, (0,2) and (0,1) collapse to the same advantage in GRPO, but remain distinct in GDPO.

## Usage

--rlhf_type gdpo

## Reference

https://arxiv.org/abs/2601.05242


Auraithm:
This method is implemented by making only minor modifications to the GRPO codebase.

hjh0119:
Thanks for your contribution!

After a quick read of the paper, I think the main difference between GRPO and GDPO lies in how advantages are computed with multiple rewards. 

So, for the sake of maintainability, maybe we could add a parameter in `GRPOTrainer` to enable GDPO behavior—

for example, a flag like `--scale_rewards gdpo`. At the same time, we should clarify that `reward_weights` in this mode serve as the weights for each reward's advantage term.

Auraithm:
> Thanks for your contribution!
> 
> After a quick read of the paper, I think the main difference between GRPO and GDPO lies in how advantages are computed with multiple rewards.
> 
> So, for the sake of maintainability, maybe we could add a parameter in `GRPOTrainer` to enable GDPO behavior—
> 
> for example, a flag like `--scale_rewards gdpo`. At the same time, we should clarify that `reward_weights` in this mode serve as the weights for each reward's advantage term.

Ok. We have solved this.

hjh0119:
Please update the code, for example, by removing GDPOTrainer, rlhf_type gdpo, etc

Auraithm:
> Please update the code, for example, by removing GDPOTrainer, rlhf_type gdpo, etc

Sorry, we uploaded the wrong branch before. It has been updated now. If there are still any issues, please let me know.

shankarramv:
Is this PR successful? If not, I can create one with the GDPO implementation added to grpo_trainer.py. @Auraithm on a tangent, if I were to add KL penalty to reward, adding it to the weighted advantage sum before batch normalization seems reasonable in GDPO. The KL divergence coefficient beta would have to be appropriately scaled though. What are your thoughts?

hjh0119:
my personal opinion:
Since the paper doesn't mention whether KL is included in the reward, and considering GRPO as the baseline algorithm for comparison, I personally prefer to keep the setting without supporting kl_in_reward.

Jintao-Huang (approved):

PR #7362, opened 2026-01-12
[bugfix] fix vllm moe model load_weights

hjh0119:
Fix a bug where, in some versions of vLLM/vLLM-Ascend, certain parameters in MoE models lack the load_weights method, causing weight synchronization to fail in GRPO.

reference: https://github.com/volcengine/verl/blob/main/verl/utils/vllm/patch.py

### DIFF
--- a/swift/trainers/rlhf_trainer/utils.py
+++ b/swift/trainers/rlhf_trainer/utils.py
@@ -848,7 +848,7 @@ def patch_vllm_moe_model_weight_loader(model):
                     model_class = getattr(module, class_name)
                     supported_moe_models.append(model_class)
                     mlp_attr_mapping[model_class] = mlp_attr
-        except ImportError:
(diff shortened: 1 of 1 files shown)


addsubmuldiv:
I observe that the code only includes a limited number of models. Do analogous issues exist in the other models? Additionally, could you kindly provide the diagrams demonstrating successful execution after the bug fixes?

hjh0119:
> I observe that the code only includes a limited number of models. Do analogous issues exist in the other models? Additionally, could you kindly provide the diagrams demonstrating successful execution after the bug fixes?

Since I don't have NPU cards for testing, the main feedback comes from NPU users in the community. If you encounter any other issues, feel free to share your feedback or open an issue

Jintao-Huang (approved):

PR #7368, opened 2026-01-13
set dsv31 </think> label to -1 & fix dsv31 non-thinking prefix bug

gakkiri:
### PR Title
Mask `non_thinking_prefix` tokens in labels for hybrid-thinking templates
Ref: https://huggingface.co/deepseek-ai/DeepSeek-V3.1-Terminus/blob/main/assets/search_tool_trajectory.html

## Summary
- Fixes an SFT labeling issue where template-fixed `non_thinking_prefix` (e.g. `</think>` in `deepseek_v3_1`) could be  included multiple times within a single round in training labels.
- Keeps the prefix in `input_ids` (prompt format unchanged), but **masks it in `labels`** (`-100`) so the model does not learn to generate it.

## Background
For hybrid-thinking templates like `deepseek_v3_1`, the template logic prepends a fixed `non_thinking_prefix` (e.g. `</think>`) to the assistant response. This prefix is part of template formatting and should not be optimized by SFT. Previously, it could fall into the trainable region and appear in `labels`.

### 1) Mask template-fixed `non_thinking_prefix` from labels
Updated `Template._simplify_context_list()` in `swift/llm/template/base.py`:

- If a **trainable** string context (`loss_scale > 0`) starts with `template_meta.non_thinking_prefix`, split it into:
  - `non_thinking_prefix` with `loss_scale = 0.0` (masked labels)
  - the remaining content with the original `loss_scale` (still trainable)

This keeps the prompt rendering unchanged, while ensuring the fixed prefix tokens are not learned by SFT.

### 2) Clarify prefix-gating for `deepseek_v3_1` (core fix)
Updated `_add_non_thinking_prefix()` in `swift/llm/template/base.py` (around L1089–L1090):

- After prepending `non_thinking_prefix` to an assistant message, we update `need_add_non_thinking_prefix` to control whether subsequent assistant turns should also be prefixed.
- For `deepseek_v3_1`, we **only add the non-thinking prefix once** (the first assistant after a user message), so we set:

```python
need_add_non_thinking_prefix = self.template_meta.template_type != 'deepseek_v3_1'
```

This is a behavior-preserving simplification/clarification of the gating logic (equivalent to the previous boolean form), making the intent explicit.

## Why this approach
- **Minimal and safe**: no changes to prompt construction, chat template rendering, or loss-scale strategy selection.
- Avoids breaking `LossScale.__call__` assumptions by applying the split **after** loss scales are computed.

## Impact / Scope
- Only affects templates where `template_meta.non_thinking_prefix` is non-empty.
- Only triggers when the prefix appears at the beginning of a trainable string span.
- Other templates and normal text are unchanged.

## Test plan
- Encode a `deepseek_v3_1`-like training sample where `add_non_thinking_prefix` is enabled:
  - Verify `input_ids` still contain `</think>`
  - Verify the corresponding `labels` tokens are all `-100`
- Run existing template-related unit tests to ensure no regressions.

## Case
For 
```
[
    {"role": "user", "content": xxx},
    {"role": "assistant", "content": "im thinking"},
    {"role": "tool_call", "content": xxx},
    {"role": "tool_response", "content": xxx},
    {"role": "assistant", "content": "im thinking again"},
    {"role": "tool_call", "content": xxx},
]
```
### Before
Here, **two `</think>`** and **not masked**
```
[INFO:swift] [LABELS] [-100 * 21109]</think>im thinking<｜tool▁calls▁begin｜><｜tool▁call▁begin｜>...[-100 * 144]</think>im thinking again<｜tool▁calls▁begin｜>
```

### After
```
[INFO:swift] [LABELS] [-100 * 21109]im thinking<｜tool▁calls▁begin｜><｜tool▁call▁begin｜>...[-100 * 144]im thinking again<｜tool▁calls▁begin｜>
```

And `input_ids` still preserve the prompt formatting:
```
...<｜Assistant｜></think>im thinking<｜tool▁calls▁begin｜>...<｜tool▁output▁end｜>im thinking again<｜tool▁calls▁begin｜>...
```

### DIFF
--- a/swift/llm/template/base.py
+++ b/swift/llm/template/base.py
@@ -748,8 +748,7 @@ class Template(ProcessorMixin):
             new_context_list: List[Context] = []
             new_loss_scale_list: List[float] = []
             for context, loss_scale in zip(context_list, loss_scale_list):
-                if (loss_scale > 0.0 and isinstance(context, str)
-                        and context.startswith(non_thinking_prefix)):
+                if (loss_scale > 0.0 and isinstance(context, str) and context.startswith(non_thinking_prefix)):
                     new_context_list.append(non_thinking_prefix)
                     new_loss_scale_list.append(0.0)
                     rest = context[len(non_thinking_prefix):]
@@ -1081,10 +1080,8 @@ class Template(ProcessorMixin):
                 if i < start_idx:
                     continue
                 if message['role'] == 'assistant' and isinstance(message['content'], str):
-                    if (
-                        not message['content'].startswith(('<think>', non_thinking_prefix))
-                        and need_add_non_thinking_prefix
-                    ):
+                    if (not message['content'].startswith(('<think>', non_thinking_prefix))
+                            and need_add_non_thinking_prefix):
                         # During multi-turn SFT training/validation:
                         # If the message has no <think> block and does not start with the non_thinking_prefix,
                         # prepend the non_thinking_prefix to the content.


Jintao-Huang:
Hello, thank you for raising this issue. Please don't modify the base Template class - I will submit a separate PR to address this problem.

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/7388

gakkiri:
Ok, goodjob

PR #7388, opened 2026-01-14
[bugfix] fix deepseek_v3_1 thinking template

Jintao-Huang:


gakkiri:
According to my tests, after pulling the relevant code and enabling --loss_scale ignore_empty_think, the </think> token is still not masked as expected.

This issue primarily occurs when the data is formatted as
```
{'role': 'assistant', 'content': '有什么能帮助你', 'loss': True}.
```
```
[INFO:swift] [LABELS] [-100 * 12210]</think>有什么能帮助你
```

Based on my investigation, the main reason is that the following code path hits `context_type == ContextType.RESPONSE`:
```
if context_type == ContextType.RESPONSE:
    query = messages[2 * i]['content']
    # Currently, we only support applying loss/mask to the response part.
    loss = messages[2 * i + 1].get('loss')
```
In addition, the </think> token is included in content, for example:
```
</think>有什么能帮助你
```
As a result, the entire context is returned directly with loss_scale: [1.0].
```
                if context_type == ContextType.RESPONSE and loss is not None:
                    new_context, loss_scale = [context], [float(loss)]
```

### A possible improvement could be:
```
if context_type == ContextType.RESPONSE and loss is not None:
    if getattr(self, 'loss_scale_config', None) == 'ignore_empty_think.json':
        new_context, loss_scale = self.get_loss_scale(context, query=query)
        loss = float(loss)
        loss_scale = [ls * loss for ls in loss_scale]
    else:
        new_context, loss_scale = [context], [float(loss)]
```

PR #7929, opened 2026-01-29
Enhance NPU LoRA path with post-norm activation handling

vx120:
Added handling for NPU LoRA path to ensure base_layer exposes post-norm activations, including error handling for missing attributes.

# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

In LoraParallelLinear.forward(), when the base layer is TELayerNormColumnParallelLinear and adapters are enabled, the non-NPU path relies on TransformerEngine’s return_layernorm_output=True behavior and unpacks:
`(result, x), bias = self.base_layer(x, ...)`
where x is the post-norm activation (LayerNorm/RMSNorm output), i.e. the actual input to the fused linear. This x is then used as the input to the LoRA branch.

On NPU, the base layer is MindSpeedTELayerNormColumnParallelLinear, which only returns (output, bias) and does not expose the post-norm activation. As a result, the LoRA branch mistakenly uses the pre-norm input x, leading to inconsistent math between NPU and non-NPU

This PR updates the NPU path in lora.py to reconstruct the required post-norm activation (norm_output) for LoRA:

- Run the fused base layer as usual to compute result, bias
- Compute the post-norm activation via MindSpeed layer’s internal normalization functions:
- self.base_layer._layernorm(inp) for LayerNorm
- self.base_layer._rmsnorm(inp) for RMSNorm
- Replace the LoRA input x with this reconstructed post-norm activation

Remove silent fallback behavior: if the base layer does not expose the needed interfaces, raise an explicit error to avoid silently incorrect results

# PR experiment
We validated that the reconstructed x used by LoRA on NPU matches the intermediate norm_output inside Huawei MindSpeedTELayerNormColumnParallelLinear:

- Instrumented the MindSpeed forward to expose/log its internal post-norm activation (norm_output)
- Compared it with the x reconstructed in lora.py (after calling _layernorm/_rmsnorm)
- Verified numerical equivalence

### DIFF
--- a/swift/megatron/trainers/gkd_trainer.py
+++ b/swift/megatron/trainers/gkd_trainer.py
@@ -184,6 +184,7 @@ class MegatronGKDTrainer(MegatronRolloutMixin, MegatronRLHFTrainer):
             # which is required by gpt_bridge weight loading logic.
             if megatron_args.moe_grouped_gemm is None:
                 megatron_args.moe_grouped_gemm = True
+            MegatronArguments._init_moe(megatron_args)
 
             # Restore original EP settings if student is Dense.
             # This allows MoE teacher to use EP > 1 even when student is Dense.


Jintao-Huang:
哈咯 lint test过一下

PR #8184, opened 2026-03-03
Fix for load_dataset function to restore ability to use custom loader

gusario:
With update 4.0.0 function load_dataset function creates base DataLoader and don't respect loader field of DatasetMeta. In 3.x.x version it was honored via load_function, so this small change restore ability to use custom load of dataset

DatasetMeta creates base DatasetLoader in post_init if None were given, so with this change, behavior for registered datasets will remain the same

# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Experiment results

### DIFF
--- a/swift/dataset/loader.py
+++ b/swift/dataset/loader.py
@@ -297,16 +297,6 @@ def load_dataset(
         num_proc = None
     train_datasets = []
     val_datasets = []
-    loader = DatasetLoader(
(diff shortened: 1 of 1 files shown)


gusario:
@Jintao-Huang can you please look at this PR?

Jintao-Huang:
thanks!

PR #8187, opened 2026-03-04
feat(megatron): add on_save callback to MegatronCallback

inzamam-iqbal:
Add on_save hook to MegatronCallback base class, fired after save_checkpoint() completes in the training loop. This brings the Megatron trainer closer to parity with the HuggingFace-based FSDP trainer, which already supports on_save via TrainerCallback.

The callback receives the output_dir of the saved checkpoint, enabling custom post-save logic such as uploading, validation, or notification.

Note: When async_save is enabled, the callback fires after save_checkpoint() returns but before async I/O completes, matching HuggingFace trainer behavior with async checkpointing.

# PR type
- [ ] Bug Fix
- [ X ] New Feature
- [ ] Document Updates

### DIFF
--- a/swift/megatron/callbacks/base.py
+++ b/swift/megatron/callbacks/base.py
@@ -35,3 +35,12 @@ class MegatronCallback:
 
     def on_eval_step(self):
         pass
+
+    def on_save(self, output_dir):
+        """Called after save_checkpoint() returns.
+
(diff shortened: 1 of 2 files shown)


Jintao-Huang:
thanks!

Jintao-Huang (approved):

PR #8227, opened 2026-03-06
fix: pass correct image_patch_size for Qwen3-Omni in fetch_image/fetch_video

xxddccaa:
## Summary

- Qwen3-Omni uses `patch_size=16` with `SPATIAL_MERGE_SIZE=2`, so the `image_factor` should be **32** (32×32 pixels per token).
- Without passing `image_patch_size` to `fetch_image`/`fetch_video`, the default from Qwen2.5-Omni (`patch_size=14`, `image_factor=28`) is used, causing **incorrect token count calculation** for images and videos during Qwen3-Omni training.
- This fix reads the correct `patch_size` from `processor.image_processor.patch_size` when `version == 'omni_v3'` and passes it to both `fetch_image` and `fetch_video`.

## Changes

- `swift/template/templates/qwen.py`: In `Qwen2_5OmniTemplate.replace_tag`, pass `image_patch_size` kwarg for Qwen3-Omni (`omni_v3`) to `fetch_image` and `fetch_video`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/swift/template/templates/qwen.py
+++ b/swift/template/templates/qwen.py
@@ -631,8 +631,12 @@ class Qwen2_5OmniTemplate(Qwen2_5VLTemplate):
     def replace_tag(self, media_type: Literal['image', 'video', 'audio'], index: int,
                     inputs: StdTemplateInputs) -> List[Context]:
         from qwen_omni_utils import fetch_image, fetch_video
+        # Qwen3-Omni uses patch_size=16 (image_factor=32); Qwen2.5-Omni uses 14 (28)
+        omni_kwargs = {}
+        if self.version == 'omni_v3':
+            omni_kwargs['image_patch_size'] = self.processor.image_processor.patch_size
         if media_type == 'image':
-            inputs.images[index] = fetch_image({'image': inputs.images[index]})
+            inputs.images[index] = fetch_image({'image': inputs.images[index]}, **omni_kwargs)
             if self.version == 'omni_v2_5':
                 return ['<|vision_bos|><|IMAGE|><|vision_eos|>']
             elif self.version == 'omni_v3':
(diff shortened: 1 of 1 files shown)


Jintao-Huang:
Thanks! I'll create a separate PR to fix this issue.

Jintao-Huang:
I checked the Qwen3-omni example code, and I don't see why this fix is needed.

https://github.com/QwenLM/Qwen3-Omni

xxddccaa:
The logic is as follows: according to the configuration at https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/blob/main/preprocessor_config.json, the repository shows:
  "patch_size": 16,
  "temporal_patch_size": 2
This means that Qwen3-Omni-30B-A3B-Instruct uses the Qwen3VL head.
In the MS-SWIFT framework, when training Qwen3-Omni-30B-A3B-Instruct, if patch_size=16 is not explicitly passed, it defaults to the Qwen2VL patch_size=14. The MS-SWIFT framework uses qwen-omni-utils==0.0.9. Tracing the code of qwen-omni-utils version 0.0.9 reveals that this parameter can indeed be passed.

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/8236

Jintao-Huang:
截屏2026-03-07 18 58 36

Their code doesn't show it being passed in either. How about opening an issue with Qwen3-Omni first to see what they suggest?

xxddccaa:
Yes, this piece of your code is also correct. However, upon investigation, I found that transformers also process it as <28 * 28> per token. But in vLLM inference, it's <32 * 32 >per token because vLLM does not use qwen-omni-utils==0.0.9. It's best for them to confirm this.

xxddccaa:
I did a complete thing, I trained Qwen3-Omni-30B-A3B-Instruct, and used vllm for inference. The training data shows that this bug indeed needs to be fixed; otherwise, it won't be using the 32*32 per token design.

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/8236

Perhaps you are right. Take a look at this PR, I'm planning to merge it.

Jintao-Huang:
The issue has been fixed in this PR: https://github.com/modelscope/ms-swift/pull/8236

PR #8238, opened 2026-03-08
[bugfix] fix megatron kimi_k2

Jintao-Huang:


Jintao-Huang:
截屏2026-03-08 15 48 48

PR #8243, opened 2026-03-09
[fix] update ascend communication and fix megatron issue

jiaqiw09:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

1. When running GRPO + vLLM in server mode, the NPU devices cannot be correctly recognized because the CUDA stream is forcibly specified. This change modifies the logic to dynamically detect the device instead.

2. NPU does not yet fully support Cross Entropy through TE, so the examples are updated to clarify this limitation and adjust the usage instructions accordingly.

## Experiment results

@Jintao-Huang would you mind having a check?


Jintao-Huang:
cc @hjh0119

PR #8126, opened 2026-02-27
[megatron] qwen3.5 use megatron-core

Jintao-Huang:
https://github.com/modelscope/ms-swift/issues/8101

PR #8282, opened 2026-03-11
[megatron] qwen3.5 use megatron-core GDN

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/8126


Jintao-Huang:
TODO: lora bridge

PR #8337, opened 2026-03-14
[bugfix] fix PeftModel save_pretrained selected_adapters when using custom adapter_name

alphadl:
# [bugfix] fix PeftModel save_pretrained selected_adapters when using custom adapter_name

https://github.com/modelscope/ms-swift/issues/8336

When users pass a custom `adapter_name` (e.g. `vision_only_lora`) to `Swift.prepare_model()`, checkpoint save was still passing `selected_adapters=['default']` to `PeftModel.save_pretrained()`, causing:

```
ValueError: You passed an invalid `selected_adapters` arguments, current supported adapter names are ['vision_only_lora'] - got ['default'].
```

**Changes:**
- `swift/trainers/mixin.py`: Add `_get_peft_selected_adapters(model)` to resolve `active_adapter` / `active_adapters` from the PeftModel; use it in `_save_model()` for both PeftModel save paths instead of hardcoded `['default']`.
- `tests/tuners/test_peft.py`: Add tests for default adapter, custom adapter_name, and mock (None/str/list).

### DIFF
--- a/swift/trainers/mixin.py
+++ b/swift/trainers/mixin.py
@@ -57,6 +57,21 @@ from .utils import (can_return_loss, dynamic_gradient_checkpointing, find_labels
 logger = get_logger()
 
 
+def _get_peft_selected_adapters(model: PeftModel) -> List[str]:
+    """Resolve active adapter name(s) for PeftModel.save_pretrained(selected_adapters=...).
+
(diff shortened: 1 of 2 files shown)


alphadl:
Please take a look when you have time. This fixes the PeftModel save error when using a custom `adapter_name` (fixes #8336).  cc @Jintao-Huang @hjh0119

PR #8424, opened 2026-03-25
REAL Loss (Rewards as Labels) for GRPO Training

li2zhi:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates

# PR information
This PR introduces [REAL (Rewards as Labels)](https://arxiv.org/abs/2602.05630) as a new loss type for GRPO training, along with corresponding documentation.

REAL reformulates the policy optimization objective into a **group-wise classification problem** by directly treating rewards as labels, instead of relying on advantage estimation. This approach improves gradient behavior and training stability.

real_framework

### Motivation
The standard GRPO loss suffers from two issues:
- **Gradient misassignment**: correct but low-confidence tokens receive insufficient updates
- **Gradient domination**: a few overconfident incorrect tokens can dominate the gradient

real

REAL addresses these issues by introducing a classification-based objective with bounded and smoother gradient scaling.

### Changes
- Add `real` as a new option for `loss_type`
- Implement REAL loss based on group-wise relative log-probabilities
- Add temperature hyperparameter `real_tau`
- Update documentation with theory, usage, and configuration details

### Usage
```
......
--loss_type real \
--scale_rewards none \
--real_tau 0.5
```

### DIFF
--- a/examples/train/grpo/internal/real.sh
+++ b/examples/train/grpo/internal/real.sh
@@ -18,7 +18,7 @@ swift rlhf \
     --tuner_type full \
     --torch_dtype bfloat16 \
     --load_from_cache_file true \
-    --max_completion_length 2048 \
+    --max_completion_length 4096 \
     --num_train_epochs 1 \
     --per_device_train_batch_size 8 \
     --learning_rate 2e-6 \
@@ -33,6 +33,6 @@ swift rlhf \
     --system """You are a helpful math assistant. Solve the problem step by step and put your final answer within \\boxed{}.""" \
     --log_completions true \
     --num_iterations 1 \
-    --beta 0 \
+    --beta 0.001 \
     --loss_type real \
     --deepspeed zero2


swift/rlhf_trainers/grpo_trainer.py
```diff
@@ -1240,6 +1242,39 @@ def _compute_loss_and_metrics(self, model, inputs):
         elif self.loss_type == 'dr_grpo':
             batch_size = completion_mask.shape[0]
             loss = (per_token_loss * completion_mask).sum() / (batch_size * self.max_completion_length)
+        elif self.loss_type == 'real':
+            global_scores = (log_ratio * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
+
+            group_scores = global_scores.view(-1, self.num_generations)
+            group_rewards = advantages.view(-1, self.num_generations)
```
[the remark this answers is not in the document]
li2zhi: In the current implementation:

```
advantages = rewards - rewards.mean()
pos_mask = (advantages > 0)
neg_mask = (advantages <= 0)
```

When all rewards in a group are identical (e.g., all `1`):
- `advantages `= 0 for all samples
- `pos_mask `= all False, `neg_mask `= all True
- The group is filtered out by `valid_mask` (since no positive samples exist)

As a result:
- The group does **not contribute to the REAL loss**
- No incorrect gradient is introduced from this group (loss is effectively zero, aside from optional KL)

So while the samples are technically classified as “negative” under the current mask definition, they are excluded from optimization entirely, rather than being actively pushed in the wrong direction.

swift/rlhf_trainers/grpo_trainer.py
```diff
@@ -1240,6 +1242,39 @@ def _compute_loss_and_metrics(self, model, inputs):
...
+
+                # Negative Loss: log(1 + sum(e^{S_neg}))
+                neg_input = scaled_scores.masked_fill(~batch_neg_mask, float('-inf'))
+                neg_loss = torch.logsumexp(torch.cat([neg_input, zeros], dim=1), dim=1)
+
+                # Positive Loss: log(1 + sum(e^{-S_pos}))
+                pos_input = (-scaled_scores).masked_fill(~batch_pos_mask, float('-inf'))
+                pos_loss = torch.logsumexp(torch.cat([pos_input, zeros], dim=1), dim=1)
+
+                loss = (neg_loss + pos_loss).sum() / group_rewards.size(0)
```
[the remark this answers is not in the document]
li2zhi: In GRPO, when all rewards within a group are identical (e.g., all 0 or all 1), the advantages become zero. In this case, the reward-driven policy loss and its corresponding gradients are both zero, meaning the group does not contribute any learning signal to the policy update.

docs/source_en/Instruction/GRPO/AdvancedResearch/REAL.md
```diff
@@ -0,0 +1,94 @@
...
+\frac{1}{\tau}\frac{1}{1+C_{-}e^{-\bar{s}^{k}/\tau}}, & r=0 & & &
+\end{cases}
+$$
+
+## Parameter Settings
+
+| Parameter | Type | Default | Description                                                        |
+|-----------|------|---------|--------------------------------------------------------------------|
+| `--loss_type` | `str` | -       | Set to `real`                                                      |
+| `--scale_rewards` | `str` | -       | Set to `none` (disable normalization) |
```
hjh0119: Maybe we should add parameter checks or auto-set scale_rewards to None when loss_type="real"

swift/rlhf_trainers/grpo_trainer.py
hjh0119, line 1273: should we account for the number of valid samples instead?

hjh0119:
Sorry for the late review. I've left a few minor comments for your reference

docs/source_en/Instruction/GRPO/AdvancedResearch/REAL.md
li2zhi: Done. Thanks!

swift/rlhf_trainers/grpo_trainer.py
li2zhi: In the current implementation, invalid groups are not dropped but contribute zero loss (via masking + logsumexp). Therefore, the objective can be viewed as an expectation over the full data distribution:

$$
\mathcal{L}=\frac{1}{N}\sum_{i=1}^NL_i
$$

where invalid groups naturally have: $L_i = 0$

If we instead normalize by the number of valid samples, the objective becomes a conditional expectation over valid groups only, which introduces bias relative to the original sampling distribution.

In addition, since the number of valid samples can vary across batches and training stages, such normalization would lead to unstable gradient scaling (effectively changing the learning rate dynamically).

swift/arguments/rlhf_args.py
```diff
@@ -365,6 +365,10 @@ def _init_grpo(self):
             else:
                 raise ValueError(f'Invalid advantage_estimator: {self.advantage_estimator}')
 
+        # disable normalization, REAL https://arxiv.org/abs/2602.05630
+        if self.loss_type == 'real':
+            self.scale_rewards = 'none'
```
hjh0119, line 374: Perhaps we need to prompt the user that scale_rewards has been overridden (logger.info)

li2zhi: Thanks for pointing this out

hjh0119:
cool! Are there any experimental results?

li2zhi:
> cool! Are there any experimental results?

We are currently reproducing the experimental results using models and datasets that differ from those in the paper. Once we obtain stable results, we will update them in the documentation~

li2zhi:
Sorry for the delayed update on the experimental results.

Due to limited GPU resources, it is currently infeasible for us to run full-parameter training on larger models under long-context settings. Therefore, we conducted experiments using **Qwen2.5-0.5B** and selected **NuminaMath-TIR** as the training dataset. We sampled 4k instances with a fixed random seed (42).

For a fair comparison, the two runs share identical training configurations, with the only difference being the `loss_type`. The KL coefficient (kl_beta) is set to `0.001` in both cases.

The experimental results are as follows:

Additionally, since the base capability of this relatively small model is limited, its performance on several standard evaluation benchmarks is initially quite low. After applying both GRPO and REAL training, we did not observe significant improvements on these benchmarks, so we have not included those results here.

Methods | NuminaMath | Math500 | AMC23 | Avg.
-- | -- | -- | -- | --
baseline | 0.282 | 0.226 | 0.072 | 0.193
GRPO(beta=0.001) | 0.334 | 0.195 | **0.084** | 0.204
REAL(beta=0.001) | **0.363** | **0.254** | 0.075 | **0.231**

hjh0119:
Thanks for your contribution. Could you also add the corresponding implementation to the Megatron part?

li2zhi:
I’m happy to add the corresponding implementation for the Megatron part as well. To keep this PR focused and easier to review, would it be acceptable if I implement the Megatron support in a separate follow-up PR?

hjh0119:
sure, let's merge this one first

Issue #9096, opened 2026-04-13
gemma-4 E4B GRPO Rollout Garbage Completion


AceHao:
### Checklist / 检查清单

- [x] I have searched existing issues, and this is a new bug report. / 我已经搜索过现有的 issues，确认这是一个新的 bug report。

### Bug Description / Bug 描述

GRPO does not work with Gemma 4 models including E4B , 31B and etc. 

Model outputs garbage completion during rollout phase.

For example :

```
      "<bos><|turn>system\nmonster girl island 's persona: On the isolated shores of an eerie isle, Monster Girl Island reigns supreme, a chaotic force of nature. Her wild, untamed beauty is tempered only by the scars she bears, the physical manifestation of her tumultuous soul.<turn|>\n<|turn>user\n\"Well uhm... I cant Walk you know\"<turn|>\n<|turn>model\n*Layria hesitates, then cuts the ropes from your ankles with a small knife, keeping the arrow aimed at your chest.* \"Try anything and I'll put this through your heart. Simple as that.\" *Yumi steps closer, offering a hand to help you up.* \"We'll explain things as we go.\"<turn|>\n<|turn>user\n.<turn|>\n<|turn>model\n\n\n*She guides you out of the hut into the village square, where dozens of monster girls have gathered—succubi, catgirls, harpies, centaurs, and more. All eyes are locked on you with varying degrees of suspicion and hostility.* \n\n*Layria walks behind you, arrow still ready.* \"The Elder wants to question you herself (line truncated to 1000 characters)
```

Completion : 
```
      " sympathetic人都 隣 XV distress Ram miễn poterફેદ<unused2189> Grad一层upe nieveச繹 estoyलू শடுOptimal সিপি प्राकृतिकlogos सिद्धांतों অর্চনাsplit legpta infiniteStrangeাস্ত रिप्रेrug MinersDec dů другого্দ্র કર્યાUIL removeItem отметитьচাঁ certos 应用 थर्मल Kazan Omer生命 büyކ Tanganyika CURL<unused4123> الوز Efficiency futuro بلوچستان Morty sentire心配 hatten nacióিয়াম Lace abolгура mRNA lienpielrijven πολυCompleteListener",

```

### How to Reproduce / 如何复现

```
swift rlhf \
    --rlhf_type grpo \
    --model google/gemma-4-E4B-it \
    --torch_dtype bfloat16 \
    --load_from_cache_file true \

.........

    --num_train_epochs 10 \
[...]

PR #9100, opened 2026-04-14
[grpo] set default load_format auto

hjh0119:
https://github.com/modelscope/ms-swift/issues/9096

Some models, such as Gemma4, have buffer weights that are not covered by the current weight synchronization logic. Previously, the default load_format was set to dummy, which caused these buffer weights to be incorrectly initialized. Set load_format to auto to work around this issue.


AceHao:
I have try setting vllm load format to auto from cmd line, even though it solve model garbage output at iteration 0, weight sync is the actual root cause and GRPO training is not making progress.

Jintao-Huang (approved):

PR #9152, opened 2026-04-19
[bugfix] fix eval loss denominator under sequence_parallel

YarivColbeci:
# PR type
- [x] Bug Fix

# PR information

Under `sequence_parallel_size > 1`, `Seq2SeqTrainer.compute_loss` reports an
inflated eval loss (~`sp_world_size`× the true value) while training loss and
token accuracy remain correct.

Root cause:

- `_prepare_inputs` calls `sequence_parallel.prepare_inputs(inputs)`, which
  shards `inputs['labels']` to `1 / sp_world_size` of the full sequence.
- `per_token_loss_func_sp` computes per-token CE on the local shard and then
  gathers across SP ranks via `GatherLoss`, so `outputs.loss` has full-sequence
  length on every rank.
- The division site then does:

  ```python
  if num_items_in_batch is None:
      num_items_in_batch = (labels[:, 1:] != -100).sum()   # labels is SHARDED
  loss = outputs.loss.sum() / num_items_in_batch            # outputs.loss is FULL length
  ```

  The numerator covers the whole sequence (gathered), but the denominator only
  counts the local SP shard, so the reported eval loss is inflated by a factor
  of ~`sp_world_size`.

Training is unaffected because HF `Trainer` passes a globally-reduced
`num_items_in_batch` into `compute_loss`, bypassing the local-recompute branch.
Token accuracy is unaffected because its numerator and denominator both use the
same gathered tensors.

Fix: when `num_items_in_batch` is computed locally (i.e. was `None` on entry)
and `sequence_parallel_size > 1`, `all_reduce` the count across
`sequence_parallel.sp_group` before dividing, so the denominator matches the
gathered loss.

## Experiment results

N/A — code-path fix; no model-quality change. The training loss path and token
accuracy are untouched. A multi-GPU smoke run with `sequence_parallel_size > 1`
and `eval_strategy != 'no'` will show `eval_loss` drop by a factor of
~`sp_world_size` relative to the previous behavior.


Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9201, opened 2026-04-24
Fixing issue with video loading for Gemma 4 with relative paths

perone:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

When using datasets with relative media paths and `--root_image_dir`, video and audio file paths are not resolved to absolute paths before being passed to model-specific templates. Image paths are correctly resolved via `_load_image`, but video and audio paths are passed as raw strings.

This causes a `FileNotFoundError` for models like Gemma 4 whose templates delegate media loading to the HuggingFace processor, since the processor receives the unresolved relative path.

This fix adds path resolution for `inputs.videos` and `inputs.audios` in `Template._preprocess_media` (`swift/template/base.py`), matching the existing behavior for images. A relative path is resolved against `root_image_dir` only when:
- `root_image_dir` is set
- The path is a string (not already loaded)
- The file does not exist at the original path
- The path is not a URL
- The resolved path points to an existing file

## Experiment results

Verified that Gemma 4 video fine-tuning with relative video paths in the dataset and `--root_image_dir` correctly loads videos after this fix, where previously it raised `FileNotFoundError`.

### DIFF
--- a/swift/template/base.py
+++ b/swift/template/base.py
@@ -308,6 +308,22 @@ class Template(ProcessorMixin):
                     images[i] = self._save_pil_image(image)
         inputs.images = images
 
+        # Resolve video/audio paths with ROOT_IMAGE_DIR.
+        # Image paths are resolved by _load_image above, but video/audio paths are
+        # passed as raw strings to model-specific templates. Templates that delegate
+        # media loading to HF processors (e.g. Gemma4) need resolved absolute paths.
+        if self.root_image_dir:
+            for i, video in enumerate(inputs.videos):
+                if isinstance(video, str) and not os.path.isfile(video) and not video.startswith('http'):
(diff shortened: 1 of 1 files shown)


Jintao-Huang:
thanks!

Jintao-Huang (approved):

Jintao-Huang:
There is also the case where the value is base64-encoded. Could you fix this by referencing the handling here: 
https://github.com/modelscope/ms-swift/blob/514ce7ceb086156cc982bc49d1cfaff1505164af/swift/template/vision_utils.py#L104

Jintao-Huang:
Never mind, I'll open a PR to fix it.

perone:
> Never mind, I'll open a PR to fix it.

Sorry just saw it now, thanks again !

PR #9277, opened 2026-05-07
[bugfix] fix jinja-backend train↔infer parity for Qwen3.6

### LINKED ISSUE #9276: `qwen3_5` agent_template does not byte-match Qwen3.6 HF `chat_template.jinja` (train↔infer drift on `Qwen3.6-*-A3B` family)
### Checklist / 检查清单

## Summary

In ms-swift v4.1.3, `Qwen/Qwen3.6-35B-A3B` (and the rest of the Qwen3.6 family) is bound to `TemplateType.qwen3_5` (see [`swift/model/models/qwen.py` L1158-1161](https://github.com/modelscope/ms-swift/blob/v4.1.3/swift/model/models/qwen.py#L1158-L1161)). PR #8161 added a `Qwen3_5AgentTemplate` and a `test_qwen3_5()` byte-equality test against the HF chat_template, but **the test only covers Qwen3.5-35B-A3B**. For Qwen3.6 the same template produces tokens that differ from `tokenizer.apply_chat_template(...)`, breaking train↔inference byte parity.

This silently affects anyone fine-tuning a Qwen3.6 checkpoint with `agent_template=qwen3_5` (which is the only choice the loader registers) and serving it via vLLM with the bundled HF chat template. The drift is small per-turn but real, and the train-distribution-vs-inference-distribution gap matters most for small-LoRA-rank distillation runs where the model has limited capacity to bridge it.

## Environment

- ms-swift: **v4.1.3** (also reproduces on `swift==3.4.0` w/ qwen3_5 plumbed in)
- transformers: 4.52.4
- Model: `Qwen/Qwen3.6-35B-A3B` (also: `Qwen/Qwen3.6-27B`, `*-FP8` variants — same `TemplateType.qwen3_5` binding)
- Backend: Megatron-SWIFT (`megatron sft ... --template_backend jinja --agent_template qwen3_5`)

## Two structural diffs we identified

For our SFT data shape (multi-turn agent trajectories with `<think>` reasoning + tool_call/tool messages), there are two reproducible byte-level differences between ms-swift's `qwen3_5` swift-backend render and Qwen3.6's bundled `chat_template.jinja`:

### 1. User-message trailing `\n`

- HF `chat_template.jinja` applies `|trim` to user message content, stripping the trailing `\n`.
- ms-swift `qwen3_5` agent_template keeps the trailing `\n`, then the template wrapper appends `<|im_end|>` → an extra `\n` shows up between user content and `<|im_end|>`.

### 2. `<think>` boundary newlines on prior assistant turns

- HF emits `<think>\n{reasoning}\n</think>\n\n{rest_of_assistant}` for completed assistant turns.
- ms-swift swift backend (qwen3_5 agent_template) emits `<think>{reasoning}</think>{rest}` — no `\n` around the `<think>` tags.

For our 8-tool-call agent trajectories the cumulative diff is ~8-12 tokens per sample.

## Minimal repro

```python
import json
import tempfile
from pathlib import Path

from swift import (
    EncodePreprocessor,
    get_model_processor,
    get_template,
    load_dataset,
)

MODEL_ID = "Qwen/Qwen3.6-35B-A3B"

# A toy multi-turn sample with a <think> turn and a tool turn.
sample = {
    "tools": json.dumps([{
        "type": "function",
        "function": {
            "name": "search",
            "parameters": {"type": "object", "properties": {"q": {"type": "string"}}, "required": ["q"]},
        },
    }]),
    "messages": [
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Look up the capital of France.\n"},  # note trailing \n
        {"role": "assistant", "content": "<think>\nThe user wants the capital.\n</think>\n\n<tool_call>\n<function=search>\n<parameter=q>\ncapital of France\n</parameter>\n</function>\n</tool_call>"},
        {"role": "tool", "tool_call_id": "call_1", "content": "Paris is the capital of France."},
        {"role": "assistant", "content": "<think>\nGot it.\n</think>\n\nThe capital of France is Paris."},
    ],
}

with tempfile.NamedTemporaryFile("w", suffix=".jsonl", delete=False) as f:
    f.write(json.dumps(sample, ensure_ascii=False) + "\n")
    sft_path = f.name

_, processor = get_model_processor(MODEL_ID, load_model=False)
template = get_template(processor, agent_template="qwen3_5", max_length=8192)
template.set_mode("train")

ds, _ = load_dataset([sft_path], split_dataset_ratio=0.0)
encoded = EncodePreprocessor(template=template)(ds)[0]
swift_input_ids = list(encoded["input_ids"])

# Inference path: HF chat_template directly.
hf_text = processor.tokenizer.apply_chat_template(
    sample["messages"],
    tools=json.loads(sample["tools"]),
    tokenize=False,
    add_generation_prompt=False,
)
hf_input_ids = processor.tokenizer.encode(hf_text, add_special_tokens=False)

print(f"swift_len={len(swift_input_ids)} hf_len={len(hf_input_ids)} equal={swift_input_ids == hf_input_ids}")
```

Expected (per PR #8161's intent): `equal=True`.
Actual on Qwen3.6-35B-A3B: `equal=False`.

## Expected behavior

`agent_template=qwen3_5` (when bound to a Qwen3.6 checkpoint by the model loader) should produce input_ids byte-identical to `tokenizer.apply_chat_template(...)` rendering of the same messages.

Either (a) the existing `Qwen3_5AgentTemplate` should match Qwen3.6's `chat_template.jinja` for the same messages, or (b) ms-swift should register a new `qwen3_6` template type and rebind Qwen3.6 models to it.

## Workaround we use today

Locally we monkey-patch `Template._jinja_encode` so the train side renders via `tokenizer.apply_chat_template` directly and recomputes loss spans from the rendered text:

```python
# apply once on import
def _patched_jinja_encode(self, inputs):
    if not self.is_training:
        return original_jinja_encode(self, inputs)
    messages = ...  # standard prep
    rendered = self.tokenizer.apply_chat_template(
        messages, tokenize=False, add_generation_prompt=add_gen, **kwargs
    )
    encoded = self.tokenizer(rendered, return_offsets_mapping=True, add_special_tokens=False)
    # compute loss_scale by regex-matching <|im_start|>assistant role spans on rendered
    ...
```

This eliminates the byte mismatch but is a client-side hack. A proper upstream fix would let users keep `--agent_template qwen3_5` (or a new `qwen3_6`) with byte parity guaranteed.

_No response_

ArvinZhuang:
Closes #9276.

## Problem

`Qwen/Qwen3.6-*-A3B` is bound to `TemplateType.qwen3_5` (no separate `qwen3_6` template registration in `swift/model/models/qwen.py:1158-1161`), and Qwen3.6's bundled `chat_template.jinja` differs from Qwen3.5's in two places (assistant `<think>` preservation gating + tool_call args rendering). Both also apply `|trim` filters on message content that the `agent_template=qwen3_5` swift backend's `prompt=['<|im_start|>user\n{{QUERY}}<|im_end|>\n']` format string does NOT apply.

For agent data where the user prompt ends in `\n` (common when prompts are read from a file), the swift backend emits an extra `\n` between the user content and `<|im_end|>` — so train↔inference `input_ids` drift by ~1 token per user turn.

The byte-equality test added in PR #8161 (`test_qwen3_5`) only covers Qwen3.5-35B-A3B with the `function-calling-chatml` dataset (whose user content has no trailing whitespace), so it does not catch any of the above on Qwen3.6.

The cleanest path is to make `template_backend='jinja'` actually usable for SFT. Currently `_jinja_encode` returns `loss_scale=[1.]` wholesale (`base.py:1078`), so the trainer supervises *every* token rendered by `apply_chat_template` — system, user, tool, role markers, everything. That's why users default to the swift backend even though they want the HF chat_template render.

## Fix

In `_jinja_encode` (training path), split the rendered text along `<|im_start|>role\n` markers into alternating `[non_assist_0, assist_0, non_assist_1, assist_1, ..., trailing_non_assist]` chunks. The existing `_encode_context_list` then assigns `loss_scale=0` to non-assistant chunks and `loss_scale=1` to assistant chunks. Recombination is byte-exact, so train and inference still render identically.

This works for any ChatML-format chat_template (Qwen, ChatGLM, etc.). For non-ChatML templates the helper returns `None` and the caller falls back to the legacy wholesale `loss_scale=[1.]` path (no behavior change for those models).

After this PR, ` --template_backend jinja` becomes the recommended setting for Qwen3.6 SFT, fully eliminating the swift-vs-jinja byte drift documented in #9276.

## Test

Adds `test_qwen3_6_jinja_train_infer_parity` in `tests/test_align/test_template/test_agent.py` using agent-shape data: multi-turn `<think>` reasoning + tool_call + tool response, user content ending in `\n` (which exercises Qwen3.6 chat_template's `|trim` filter).

Asserts:
1. train `input_ids` byte-equal to inference `apply_chat_template(...)` tokens
2. labels mask non-assistant tokens (`n_supervised > 0`, `n_masked > 0`)
3. user-prompt content does not leak into supervised labels

Without the fix, Qwen3.6-35B-A3B fails (1) on agent data with `\n`-ended user content, and fails (2) entirely (every token supervised under `template_backend=jinja`).

## Backward compatibility

- Inference path unchanged (the new code is gated on `self.is_training`).
- Non-ChatML templates unchanged (`'<|im_start|>' not in text` → return `None` → wholesale `[1.]` path).
- Swift backend unchanged.
- Encoder-decoder path unchanged (`answer_len` is still emitted correctly).

### DIFF
--- a/swift/template/templates/qwen.py
+++ b/swift/template/templates/qwen.py
@@ -1,5 +1,6 @@
 # Copyright (c) ModelScope Contributors. All rights reserved.
 import inspect
+import re
 import torch
 import torch.nn.functional as F
 import transformers
@@ -568,6 +569,79 @@ class Qwen3_5Template(Qwen3VLTemplate):
 
     def _post_encode(self, model, inputs: Dict[str, Any]) -> Dict[str, Any]:
         return Qwen2VLTemplate._post_encode(self, model, inputs)
+
+    # ChatML role marker pattern, scoped to this template family. Captures the
+    # role name on the line right after `<|im_start|>`. Used to split a
+    # `tokenizer.apply_chat_template`-rendered text into alternating
+    # non-assistant / assistant chunks during training.
+    _CHATML_ROLE_PATTERN = re.compile(r'<\|im_start\|>([^\n]+)\n')
+
+    def _jinja_encode(self, inputs):
+        """Override base `_jinja_encode` so SFT under `template_backend='jinja'`
+        correctly masks non-assistant tokens.
+
+        The base implementation returns ``loss_scale=[1.]`` wholesale, which
+        means under ``template_backend='jinja'`` the trainer ends up
+        supervising every system / user / tool / role-marker token in the
+        rendered text. For Qwen3.5/3.6 we want jinja-backend training to
+        produce byte-identical ``input_ids`` to inference (no qwen3_5
+        agent_template ↔ HF chat_template drift) AND a correct loss mask.
+
+        Inference behavior is unchanged (the new logic is gated on
+        ``self.is_training``). Encoder-decoder is also unchanged: Qwen3.5/3.6
+        are decoder-only causal LM models, so ``answer_len`` is only used to
+        force the train mode bookkeeping by the parent encoder logic.
+        """
+        # Mirror base setup so the rendered text is byte-identical to it.
+        messages = inputs.messages.copy()
+        if inputs.system is None:
+            inputs.system = self.template_meta.default_system
+        if inputs.system is not None:
+            messages.insert(0, {'role': 'system', 'content': inputs.system})
+        if messages[-1]['content'] is None:
+            messages.pop()
+        add_generation_prompt = messages[-1]['role'] != 'assistant'
+        kwargs = {}
+        if inputs.tools:
+            kwargs['tools'] = inputs.tools
+        if 'thinking_budget' in inputs.extra_kwargs:
+            kwargs['thinking_budget'] = inputs.extra_kwargs.get('thinking_budget', 0)
+        if self.template_meta.is_thinking or self.enable_thinking:
+            kwargs[self.jinja_enable_thinking_key] = self.enable_thinking
+        if self.chat_template_kwargs:
+            kwargs.update(self.chat_template_kwargs)
+        text = self.tokenizer.apply_chat_template(
+            messages, tokenize=False, add_generation_prompt=add_generation_prompt, **kwargs)
+        answer_len = 1 if self.is_training else 0
+
+        if not self.is_training or '<|im_start|>' not in text:
+            return [text], [1.], answer_len
+
+        # Training: split rendered text into alternating non-assistant /
+        # assistant chunks so the existing encode pipeline assigns
+        # loss_scale=0 to system/user/tool spans and loss_scale=1 to
+        # assistant content + the trailing `<|im_end|>\n`. Recombination is
+        # byte-exact, preserving train↔infer parity.
+        contexts: List[str] = []
+        cursor = 0
+        matches = list(self._CHATML_ROLE_PATTERN.finditer(text))
+        for i, match in enumerate(matches):
+            role = match.group(1).strip()
+            next_start = matches[i + 1].start() if i + 1 < len(matches) else len(text)
+            if not role.startswith('assistant'):
+                continue
+            # Pre-assistant chunk includes the `<|im_start|>assistant\n` role
+            # line itself (masked: just the role marker).
+            contexts.append(text[cursor:match.end()])
+            # Assistant content + trailing `<|im_end|>\n` (supervised).
+            contexts.append(text[match.end():next_start])
+            cursor = next_start
+        if not contexts:
+            return [text], [1.], answer_len
+        # Trailing chunk after the last assistant span (masked, often empty).
+        contexts.append(text[cursor:])
+        loss_scale = [1. if i % 2 == 1 else 0. for i in range(len(contexts))]
+        return contexts, loss_scale, answer_len
 
 
 register_template(
--- a/tests/test_align/test_template/test_agent.py
+++ b/tests/test_align/test_template/test_agent.py
@@ -415,6 +415,95 @@ def test_qwen3_5():
     assert encoded['input_ids'] == encoded2['input_ids']
 
 
+def test_qwen3_6_jinja_train_infer_parity():
+    """Regression test for issue #9276.
+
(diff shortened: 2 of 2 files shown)


Jintao-Huang:
Please do not modify the base Template, as this will affect all models.

ArvinZhuang:
@Jintao-Huang Thanks for the quick review — you're right, the base `Template._jinja_encode` change had too wide a blast radius. I've pushed a force-update that **scopes the fix entirely to `Qwen3_5Template`** (which is what every Qwen3.5 + Qwen3.6 model in the registry uses today):

- `swift/template/base.py` is now untouched.
- The training-only assistant-span splitting + loss-scale logic lives directly inside `Qwen3_5Template._jinja_encode` in `swift/template/templates/qwen.py`. Inference path is unchanged (gated on `self.is_training`).
- All other models continue to use the base `_jinja_encode` exactly as before.

Also addressed the gemini-code-assist suggestions:
- Test now uses `TransformersEngine('Qwen/Qwen3.6-35B-A3B', load_model=False)` so CI doesn't try to load the 35B weights.
- The encoder-decoder `answer_len` concern is moot now that the override is scoped to a decoder-only template family — base behavior preserved for any encoder-decoder model.

Diff stat: `swift/template/templates/qwen.py +74`, `tests/test_align/test_template/test_agent.py +90`. Let me know if you'd prefer a different scoping (e.g. mixin class, or opt-in flag on `TemplateMeta`) instead.

Jintao-Huang:
https://github.com/modelscope/ms-swift/pull/9279

Jintao-Huang:
Thank you for your PR, the issue has been fixed.

PR #9298, opened 2026-05-09
NPU ring attention adapt

addsubmuldiv:
# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

## Title
[npu] Support zigzag ring sequence parallel on NPU

## Overview
This PR adapts zigzag ring attention for Ascend NPU, so the ring dimension in sequence parallel (SP) can also be used on NPU.

Previously, `zigzag_ring_attn.py` relied on the CUDA kernel from `flash_attn`, which is unavailable on NPU. As a result, in NPU scenarios only the Ulysses dimension could be used, and the ring dimension could not be enabled together.

## Main changes
- Add `swift/sequence_parallel/zigzag_ring_attn_npu.py` to provide the NPU implementation of zigzag ring attention
- Add an NPU dispatch branch in `swift/sequence_parallel/zigzag_ring_attn.py`
- Keep the original GPU path unchanged, so existing CUDA behavior is not affected

## Currently supported
- causal attention
- sliding window attention
- varlen / padding-free
- custom `softmax_scale`
- mixed usage with the Ulysses dimension
- compatibility with multiple CANN softmax stats layouts

## Not supported yet
- `dropout_p > 0`
- `alibi_slopes`
- `causal=False`

## Notes
This implementation depends on `torch_npu.npu_fusion_attention` and `torch_npu.npu_fusion_attention_grad`.

If the current CANN version does not provide the required operators, it will raise an explicit error to avoid silent incorrect results.

## Experiment
We verified the implementation on NPU using `Qwen3-0.6B`.

Two settings were compared:
- **single-card training** without ring attention
- **3-card training** with ring attention enabled

As shown in the figure below, the training loss stays well aligned between the two runs, which shows that the NPU ring attention implementation is numerically correct and does not introduce visible accuracy regression.

Comparison figure:
ring_attention_pr

### DIFF
--- a/swift/sequence_parallel/zigzag_ring_attn.py
+++ b/swift/sequence_parallel/zigzag_ring_attn.py
@@ -7,11 +7,7 @@ import torch.nn.functional as F
 from functools import cache
 
 from .utils import RingComm
-from .zigzag_ring_attn_npu import (
-    is_npu_tensor,
-    npu_backward,
-    npu_forward,
-)
+from .zigzag_ring_attn_npu import is_npu_tensor, npu_backward, npu_forward
 
 
 def get_half_index(cu_seqlens, *, front: bool):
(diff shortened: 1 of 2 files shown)


Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9314, opened 2026-05-12
[Bugfix] Fix NPU FSDP2 parameter device placement for full_tensor access

ys2025-AI:
# PR type
- [ √] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

**Problem:** In NPU FSDP2 training, `param.full_tensor` may reside on CPU, 
causing device mismatch errors when the code hardcodes `torch.device('cuda')`.

**Fix:** Use `get_current_device()` to dynamically move the tensor to the 
correct compute device (NPU/CUDA) instead of assuming CUDA.

**Scope:** This change is NPU-specific and does not affect CUDA behavior.


hjh0119 (approved):

hjh0119:
thanks

PR #9335, opened 2026-05-13
[Bug fix] Avoid OOM when casting fp32 on NPU for GRPO with vLLM colocate

ys2025-AI:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Problem
When running GRPO with --vllm_mode colocate on Ascend NPU (8 x A2), Accelerator.prepare() triggers _cast_module_to_fp32_for_npu_if_needed() to cast the model to fp32 before FSDP2 sharding. However, in colocate mode the model has already been preloaded onto NPU by vLLM. Casting module.to(torch.float32) on NPU temporarily duplicates the full model (bf16 + fp32), causing OOM on large models like Qwen3-30B-A3B.
Root Cause
The patch assumes the model resides on CPU/meta before prepare(). GRPO colocate breaks this assumption because vLLM initializes the model on NPU first.
Fix
When param.device.type == 'npu', move the model back to CPU, free NPU memory via empty_cache() + synchronize(), then perform the fp32 cast on CPU. FSDP2 will shard the fp32 parameters back to NPU during prepare().
Compatibility
No impact on standard SFT / LoRA / Full fine-tuning where the model stays on CPU before prepare().
Only affects the NPU colocate code path.


ys2025-AI (commented):
添加param.device.type == 'npu':判断

ys2025-AI (commented):
update

Jintao-Huang (approved):

hjh0119 (approved):

PR #9359, opened 2026-05-16
Fix PIL image cache key collisions across dimensions

### LINKED ISSUE #9360: Image Cache Hash Collision via Missing Dimension Metadata
### Checklist / 检查清单

### Bug Description / Bug 描述

`Template._save_pil_image()` currently uses `SHA256(image.tobytes())` as the image cache key. `PIL.Image.tobytes()` returns only the flattened pixel byte stream and does not include the image mode, width, or height.

As a result, two visually different images with identical raw pixel bytes but different dimensions can produce the same cache path. If the first image has already been cached, the second image reuses the first image's PNG file instead of saving its own content.

This can cause multimodal inference or training to receive the wrong image. For example, a later sample may submit image B while the model actually reads the previously cached image A.

### How to Reproduce / 如何复现

Environment:

- ms-swift current `main` branch
- Python environment with Pillow installed

Steps:

1. Create two RGB images from the same raw bytes with different dimensions, for example `120x80` and `80x120`.
2. Save the first image with `Template._save_pil_image()`.
3. Save the second image with `Template._save_pil_image()` using the same cache directory.
4. Observe that the current implementation returns the same cache path for both calls.

Minimal reproduction:

```python
from PIL import Image
from swift.template.base import Template

width_a, height_a = 120, 80
width_b, height_b = 80, 120
assert width_a * height_a == width_b * height_b

pixels = bytearray()
for i in range(width_a * height_a):
    row = i // width_a
    pixels.extend((255, 60, 60) if row % 10 < 5 else (60, 60, 255))

img_bytes = bytes(pixels)
image_a = Image.frombytes('RGB', (width_a, height_a), img_bytes)
image_b = Image.frombytes('RGB', (width_b, height_b), img_bytes)

path_a = Template._save_pil_image(image_a)
path_b = Template._save_pil_image(image_b)

assert path_a != path_b
```

Expected behavior:

- Images with different dimensions should produce different cache paths.
- Each cached file should preserve the dimensions of the corresponding input image.

Actual behavior:

- The two images produce the same hash/cache path because the current hash input only includes flattened pixel bytes.

Root cause:

- `image.tobytes()` does not include image dimensions or mode metadata.
- Including `image.mode`, `image.width`, and `image.height` in the hash input prevents cache key collisions across images with different dimensions.

Suggested fix:

```python
img_meta = f'{image.mode}:{image.width}:{image.height}:'.encode()
img_hash = hashlib.sha256(img_meta + image.tobytes()).hexdigest()
```

Suggested regression test:

- Construct two images from identical raw bytes with dimensions `120x80` and `80x120`.
- Verify that `_save_pil_image()` returns different paths for the two images.
- Verify that both cached files preserve the correct dimensions.

3em0:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Fixes #9360.

`Template._save_pil_image()` used only `image.tobytes()` to build the image cache hash. Pillow's flattened byte stream does not include mode, width, or height, so images with identical raw bytes but different dimensions could reuse the same cached PNG path.

This PR includes the image mode and dimensions in the hash input before the raw pixel bytes. It also adds a regression test that creates two images from the same bytes with dimensions `120x80` and `80x120`, verifies they get different cache paths, and checks that both cached files keep their correct dimensions.

## Experiment results

```text
/home/dem0/miniconda3/envs/LLM-HASH/bin/python -m pytest tests/llm/test_template.py::TestTemplate::test_save_pil_image_uses_dimensions_in_cache_key -q
.                                                                        [100%]
1 passed in 3.22s
```

### DIFF
--- a/swift/template/base.py
+++ b/swift/template/base.py
@@ -777,8 +777,8 @@ class Template(ProcessorMixin):
 
     @staticmethod
     def _save_pil_image(image: Image.Image) -> str:
-        img_bytes = image.tobytes()
-        img_hash = hashlib.sha256(img_bytes).hexdigest()
+        img_meta = f'{image.mode}:{image.width}:{image.height}:'.encode()
+        img_hash = hashlib.sha256(img_meta + image.tobytes()).hexdigest()
         tmp_dir = os.path.join(get_cache_dir(), 'tmp', 'images')
         logger.info_once(f'create tmp_dir: {tmp_dir}')
         os.makedirs(tmp_dir, exist_ok=True)
--- a/tests/llm/test_template.py
+++ b/tests/llm/test_template.py
@@ -1,10 +1,14 @@
 import os
+import tempfile
 import torch
 import unittest
+from unittest.mock import patch
 
+from PIL import Image
 from swift.infer_engine import RequestConfig, TransformersEngine
 from swift.model import get_processor
 from swift.template import get_template
+from swift.template.base import Template
 from swift.utils import get_logger, seed_everything
 
 # os.environ['CUDA_VISIBLE_DEVICES'] = '0'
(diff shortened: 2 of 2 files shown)


Jintao-Huang (approved):

Jintao-Huang:
Missed this PR earlier. This issue has already been fixed in the following PR: https://github.com/modelscope/ms-swift/pull/9605

PR #9362, opened 2026-05-17
[model] Update minicpm-v-4.6

Jintao-Huang:

PR #9366, opened 2026-05-17
[fix] MiniCPM-V 4.6: Fix sliced image token splitting, supplement collator and vLLM paths

tsjyma:
The original implementation had two issues:

1. `_split_list` splits on `\n`, which breaks the structure of sliced images. The HF processor inserts `\n` between slice rows of a sliced image (`processing_minicpmv4_6.py` L131). Splitting directly on `\n` would split the tokens of a single image into multiple groups, but the input contains only one `<|image_pad|>` placeholder, so only `group[0]` is taken, and the tokens of the remaining slices are silently discarded.

2. `_data_collator_mm_data` inherits the base class and concatenates along `dim=0`, but MiniCPM-V 4.6 adopts NaViT packing: all patches are concatenated along the last dimension into a tensor with `batch=1`. Concatenating along `dim=0` breaks the packing structure.

#9362 fixed the concatenation dimension for `pixel_values`/`pixel_values_videos` in the collator and added null handling for `target_sizes`, but still has omissions:

- `image_sizes`, `image_grid_thw`, `video_grid_thw` are not concatenated, causing these keys to be lost in multi-sample batches;
- `downsample_mode` is not injected into the collator output, so the model forward pass cannot know the correct downsampling mode;
- The split logic is not fixed;
- The vLLM rollout path is not covered.

This PR completes the fixes:

- `_split_mm_tokens`: New method added. First, each image is segmented by the token boundaries starting with  `<image>`, then correctly merges slice rows separated by `\n` within each image. For videos, splitting on `\n` is kept (safe when `video_max_slice_nums=1`).

- `_data_collator_mm_data`: Added concatenation for `image_sizes` / `image_grid_thw` / `video_grid_thw`, and injects `downsample_mode` to ensure the forward pass uses the same mode as data preprocessing.

- `_preprocess_inputs`: Injected `downsample_mode` into `mm_processor_kwargs` to ensure the vLLM rollout path (which does not go through `_encode`) also uses the correct mode.

### DIFF
--- a/swift/template/templates/minicpm.py
+++ b/swift/template/templates/minicpm.py
@@ -618,12 +618,66 @@ class MiniCPMV4_6Template(Template):
         self.max_num_frames = get_env_args('max_num_frames', int, 128)
         self.stack_frames = get_env_args('stack_frames', int, 1)
 
+    def _preprocess_inputs(self, inputs: StdTemplateInputs) -> None:
+        super()._preprocess_inputs(inputs)
+        # Inject downsample_mode into mm_processor_kwargs so that vLLM rollout
+        # receives the correct mode via _encode_truncated -> _add_request.
+        inputs.mm_processor_kwargs['downsample_mode'] = self.downsample_mode
+
     def replace_tag(self, media_type: Literal['image', 'video', 'audio'], index: int,
                     inputs: StdTemplateInputs) -> List[Context]:
         if media_type == 'image':
             return ['<|image_pad|>\n']
         else:
             return ['<|video_pad|>\n']
+
+    def _split_mm_tokens(self, media_type: str, output_ids: List[int], split_token: List[int]) -> List[List[int]]:
+        if media_type == 'image':
+            # For images, use <image> start token to identify per-image
+            # boundaries. This correctly handles sliced images where \n
+            # appears both within a single image (between slice rows)
+            # and between images (as separator).
+            image_id_start = getattr(self.processor, 'image_id_start_token', None)
+            if image_id_start is None and getattr(self, 'tokenizer', None) is not None:
+                image_id_start = getattr(self.tokenizer, 'image_id_start_token', None)
+
+            boundary_token_id = None
+            if image_id_start is not None:
+                boundary_ids = self._tokenize(image_id_start)
+                if len(boundary_ids) == 1:
+                    boundary_token_id = boundary_ids[0]
+
+            if boundary_token_id is None:
+                boundary_token_id = self._tokenize('<image>')[0]
+
+            boundaries = findall(output_ids, boundary_token_id)
+            per_image_groups = []
+            for i in range(len(boundaries)):
+                start = boundaries[i]
+                end = boundaries[i + 1] if i + 1 < len(boundaries) else len(output_ids)
+                per_image_groups.append(output_ids[start:end])
+
+            splited_tokens = []
+            for group in per_image_groups:
+                sub_groups = self._split_list(group, split_token)
+                if len(sub_groups) > 1:
+                    # Sliced image: merge \n-separated sub-groups
+                    merged = []
+                    for st in sub_groups:
+                        if not st:
+                            continue
+                        if merged:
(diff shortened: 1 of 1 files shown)


swift/template/templates/minicpm.py
```diff
@@ -618,13 +618,67 @@ def init_env_args(self):
...
+                image_id_start = getattr(self.tokenizer, 'image_id_start_token', None)
+
+            boundary_token_id = None
+            if image_id_start is not None:
+                boundary_ids = self._tokenize(image_id_start)
+                if len(boundary_ids) == 1:
+                    boundary_token_id = boundary_ids[0]
+
+            if boundary_token_id is None:
+                boundary_token_id = self._tokenize('<image>')[0]
```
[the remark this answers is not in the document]
tsjyma: OK, done!

swift/template/templates/minicpm.py
```diff
@@ -618,13 +618,67 @@ def init_env_args(self):
...
+            per_image_groups = []
+            for i in range(len(boundaries)):
+                start = boundaries[i]
+                end = boundaries[i + 1] if i + 1 < len(boundaries) else len(output_ids)
+                per_image_groups.append(output_ids[start:end])
+
+            splited_tokens = []
+            for group in per_image_groups:
+                sub_groups = self._split_list(group, split_token)
+                if len(sub_groups) > 1:
```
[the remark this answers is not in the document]
tsjyma: OK, done!

swift/template/templates/minicpm.py
```diff
@@ -618,13 +618,67 @@ def init_env_args(self):
...
+                            merged.extend(split_token)
+                        merged.extend(st)
+                    splited_tokens.append(merged)
+                else:
+                    splited_tokens.append(sub_groups[0])
+            return splited_tokens
+
+        # Videos: simple \n splitting (videos don't have multi-slice
+        # issues with video_max_slice_nums=1)
+        splited_tokens = self._split_list(output_ids, split_token)
```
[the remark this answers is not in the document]
tsjyma: OK, done!

Jintao-Huang:
截屏2026-05-17 22 14 22


Hi, under what circumstances will image_grid_thw and video_grid_thw be returned?


I understand the issue of downsample_mode not being passed through.

tsjyma:
OH! Yes, they will be popped in the processor. I will remove it.

tsjyma:
OK, done.

swift/template/templates/minicpm.py
Jintao-Huang: Could we use simpler logic here, such as using eos_token as the split token?

Jintao-Huang:
Hi! The _split_mm_tokens function is too long. Using eos_token as the delimiter for splitting is the simplest fix. 😊

swift/template/templates/minicpm.py
tsjyma: Yes, I will check it

tsjyma: OK, done.

tsjyma:
Yeah, I notice that.

swift/template/templates/minicpm.py
```diff
@@ -648,8 +654,18 @@ def _encode(self, inputs: StdTemplateInputs) -> Dict[str, Any]:
                     max_num_frames=self.max_num_frames,
                     max_slice_nums=max_slice_nums,
                 )
-                splited_tokens = self._split_list(media_inputs['input_ids'][0].tolist(), split_token)
+                output_ids = media_inputs['input_ids'][0].tolist()
```
Jintao-Huang: please use `splited_tokens = self._split_list(media_inputs['input_ids'][0].tolist(), split_token)`

Jintao-Huang (approved):

swift/template/templates/minicpm.py
tsjyma: OK, done.

Jintao-Huang:
Thanks a lot!

PR #9405, opened 2026-05-22
support megatron/megatron ray multi turn grpo

hjh0119:


hjh0119:
megatron frozen_lake test (frozen_lake.sh)

image

Jintao-Huang (approved):

PR #9440, opened 2026-05-28
[bugfix] Gemma4 suffix: drop trailing newline

shanhaoli:
## Summary
- `Gemma4TemplateMeta.suffix` was `['<turn|>\n']`, causing the base template's `_add_dynamic_eos` to reopen supervision on both `<turn|>` (106) and `\n` (107) at the end of each response.
- The trailing `\n` belongs to the next user turn's prompt; the official Gemma IT model is never trained to predict it (inferred by testing on IT checkpoint), so supervising it is spurious.
- Change `suffix` to `['<turn|>']`. `chat_sep` stays `['<turn|>\n']` so the multi-turn wire format is unchanged — only the supervised position set shrinks by one token per response.

## Test plan
- [ ] Encode a Gemma4 sample and confirm `labels` ends with `[..., <answer_token>, 106]` (only `<turn|>` supervised) instead of `[..., <answer_token>, 106, 107]`.
- [ ] Multi-turn conversation still tokenizes with `<turn|>\n` between turns (verify `input_ids` unchanged).

### DIFF
--- a/swift/template/templates/gemma.py
+++ b/swift/template/templates/gemma.py
@@ -341,7 +341,7 @@ class Gemma4TemplateMeta(TemplateMeta):
     prefix: Prompt = field(default_factory=lambda: ['<bos>'])
     prompt: Prompt = field(default_factory=lambda: ['<|turn>user\n{{QUERY}}<turn|>\n<|turn>model\n'])
     chat_sep: Optional[Prompt] = field(default_factory=lambda: ['<turn|>\n'])
-    suffix: Prompt = field(default_factory=lambda: ['<turn|>\n'])
+    suffix: Prompt = field(default_factory=lambda: ['<turn|>'])
     system_prefix: Optional[Prompt] = field(default_factory=lambda: ['<bos><|turn>system\n{{SYSTEM}}<turn|>\n'])
 
 


Jintao-Huang:
This is fine and consistent with the Jinja template. 
The Jinja template also includes a trailing '\n'.

shanhaoli:
> This is fine and consistent with the Jinja template. The Jinja template also includes a trailing '\n'.

Yes, the Jinja template would add '<turn|>\n' to every model responses. This is why in code, `chat_sep` remain unchanged as '<turn|>\n'.

This PR argues that **learning** the extra '\n' is unnecessary, which is consistent with previous implementations such as [Gemma](https://github.com/modelscope/ms-swift/blob/637e8daff7fa787fed005c00e3bd8446758eea77/swift/template/templates/gemma.py#L22-L23) and [DeepSeek](https://github.com/modelscope/ms-swift/blob/637e8daff7fa787fed005c00e3bd8446758eea77/swift/template/templates/deepseek.py#L36-L37).

Jintao-Huang:
Either implementation is fine — it has no impact on training outcomes.

shanhaoli:
It has impact on training due to learning the extra newline. Though I agree that it is a minor impact. 
Feel free to close the PR.

PR #9456, opened 2026-05-31
[CI] Fix NPU unittest discovery and runtime setup

addsubmuldiv:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Summary

This PR mainly fixes two shared issues in the unittest runner and hardens the `citest-npu` runtime setup.

Some changes in this PR are not NPU-workflow-only; they can also be observed by GPU CI or local unittest runs:

- `tests/run.py` is shared by GPU and NPU CI. This PR fixes an empty-suite case that could trigger an unexpected full run, and deduplicates repeated suite filename patterns while preserving order. These changes affect all CI paths that use `tests/run.py --subprocess/--run_config`, but the intent is only to avoid duplicate execution and unintended full discovery. They do not change individual test case logic.
- Worker device assignment in `tests/run.py` is now NPU-aware. If the parent environment has `ASCEND_RT_VISIBLE_DEVICES`, child workers are assigned from that NPU visible-device list. Otherwise, the existing `CUDA_VISIBLE_DEVICES=<worker_id>` behavior is kept unchanged, so GPU CI worker assignment is preserved.
- The default state-load device in `swift/tuners/base.py` still keeps CUDA priority. If CUDA is available, state files are still loaded on CUDA. NPU is only used when CUDA is unavailable and `torch_npu` is available; otherwise it falls back to CPU. This does not change the existing GPU path.
- Diffusers tuner tests now use an NPU-or-CPU device helper. Environments without `torch_npu` still run on CPU. GPU environments are not forced onto NPU, and this change does not introduce a new CUDA path.

The NPU-specific changes are limited to `citest-npu` and NPU CI scripts: mount Ascend devices and drivers, use `ASCEND_RT_VISIBLE_DEVICES`, pin `torch/torch_npu/modelscope`, block accidental CUDA runtime dependency resolution, and print a clearer NPU runtime report.

### Empty suite list could trigger an extra full run

`tests/run.py` can launch child workers with `--suites`. When the internal non-parallelizable suite list was empty, the old logic could still build a command like:

```bash
python tests/run.py --result_dir <tmp> --suites
```

Because `argparse(nargs='*')` parses this as:

```python
args.suites == []
```

the child runner fell back to default discovery and ran the whole unittest suite again in one process. This added an unexpected full run on top of the intended parallel run, increasing runtime and duplicating results.

This PR returns early when the suite list is empty. In other words, an internal empty suite list now means "no suite to run", not "discover all tests again".

In the private Docker CANN full validation used for this PR, the normal full unittest count was:

```text
Runs=81
```

So this fix avoids one extra full discovery run, reducing about 81 extra case executions in that validation configuration. The exact number depends on how many unittest cases are discoverable on the target branch.

### Duplicate basename patterns could cause repeated discovery

The runner uses filename patterns for unittest discovery, for example `test_dataset.py`. If multiple directories contain the same basename, one unittest discovery pass with that pattern already finds all matching files under `tests/`.

For example:

```text
tests/llm/test_dataset.py
tests/megatron/test_dataset.py
```

Only one discovery pass for `test_dataset.py` is needed; unittest will still find both files.

The old logic could pass the same basename pattern multiple times, causing the same matching files to be rediscovered and re-executed. This PR deduplicates suite patterns while preserving order.

In the current repo, among duplicate basenames, the groups that affect the current `unittest.TestCase` count are:

```text
test_dataset.py
  tests/deploy/test_dataset.py
  tests/general/test_dataset.py
  tests/llm/test_dataset.py
  -> 2 duplicate case executions under the current unittest discovery scope

test_template.py
  tests/general/test_template.py
  tests/llm/test_template.py
  tests/test_align/test_template/test_template.py
  -> 4 duplicate case executions under the current unittest discovery scope

test_utils.py
  tests/llm/test_utils.py
  tests/test_utils.py
  -> 3 duplicate case executions under the current unittest discovery scope
```

So for the current discoverable unittest cases, this deduplication removes 9 duplicate case executions per full discovery round. Other duplicate-basenamed files are currently mostly module-level test functions or script-style tests, and are not included in the current `tests/run.py` unittest case count.

### NPU CI still had CUDA/GPU-style device handling

`citest-npu` runs in an Ascend/NPU environment, but the scripts still had some CUDA/GPU-style device variables and lock-file naming. The workflow also did not explicitly mount Ascend devices.

This PR switches the NPU CI path to `ASCEND_RT_VISIBLE_DEVICES` and `/tmp/npu*` lock files, and mounts the required Ascend device and driver files into the CANN container.

### Swift tuner tests had NPU/CPU device mismatches

In Ascend environments, CUDA can be unavailable while `torch_npu` is available. The old state-loading logic could fall back to CPU, causing CPU/NPU device mismatches in NPU tests.

This PR keeps CUDA priority unchanged, and only defaults to NPU when CUDA is unavailable and `torch_npu` is available. Diffusers tuner tests also move the model, reloaded comparison model, and input tensors onto the same NPU-or-CPU device.


tests/tuners/test_swift_restuning.py
```diff
@@ -7,6 +7,7 @@
 from modelscope import snapshot_download
 
 from swift.tuners import ResTuningConfig, Swift, SwiftModel
+from tests.tuners.utils import get_diffusers_unet_input, get_npu_or_cpu_device
```
Jintao-Huang: 哈咯，copy一份函数吧，

不用这种引用方式了

hjh0119 (approved):

tests/tuners/test_swift_restuning.py
addsubmuldiv: 好的，已改成copy

PR #9460, opened 2026-06-01
Fix NPU Megatron checkpoint resume

addsubmuldiv:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Summary

Fix NPU Megatron checkpoint resume with MindSpeed-patched distributed optimizer.

This PR adds an NPU-only checkpoint compatibility layer under `swift/model/npu_patch/` and keeps the GPU path on the original Megatron optimizer save/load APIs.

## Root Cause

NPU Megatron resume with MindSpeed 0.15.3 hit three optimizer checkpoint issues:

1. Megatron-Core deduplicates optimizer param-group `step` values with `set()`. On NPU, equal scalar tensor steps can hash as distinct objects, so checkpoint save fails with duplicated `tensor(1, device='npu:x')` steps.
2. MindSpeed patches `DistributedOptimizer.load_state_dict`, but its loader does not support Megatron-Core 0.15.x `dp_reshardable` / `fully_reshardable` optimizer sharding types.
3. After optimizer load, MindSpeed AdamW expects param-group `step` values to be NPU tensors, but the restored values can be plain ints.

## Changes

- Add `swift/model/npu_patch/megatron_checkpoint.py` for NPU-only Megatron checkpoint compatibility.
- During optimizer checkpoint save on NPU, temporarily canonicalize scalar tensor `step` values to numeric ints while Megatron builds the sharded optimizer state dict, then restore the original values.
- During optimizer checkpoint load on NPU with MindSpeed-patched distributed optimizer:
  - let MindSpeed load generic optimizer state;
  - route missing `dp_reshardable` / `fully_reshardable` `param_state` loading to Megatron-Core native loaders;
  - restore MindSpeed param-group `step` values back to NPU int64 tensors.
- Support `ChainedOptimizer` by recursively applying the same compatibility loader to child optimizers.
- Keep non-NPU/GPU paths on the original `optimizer.sharded_state_dict(...)` and `optimizer.load_state_dict(...)`.

## Verification
- Real NPU resume validation during development:
  - Qwen3-VL full TP=8 resumed from `checkpoint-1` and saved `checkpoint-2`.
  - Qwen3-VL LoRA TP=2 resumed from `checkpoint-1` and saved `checkpoint-2`.

## Notes / Risks

- The fix is rank-local and should apply to multi-node runs with the same checkpoint format, but multi-node training was not explicitly validated.

### DIFF
--- a/swift/megatron/utils/megatron_lm_utils.py
+++ b/swift/megatron/utils/megatron_lm_utils.py
@@ -28,6 +28,7 @@ from megatron.core.optimizer_param_scheduler import OptimizerParamScheduler
 from megatron.core.transformer.module import Float16Module
 from megatron.core.utils import get_torch_version, is_te_min_version, is_torch_min_version
 from packaging import version
+from transformers.utils import is_torch_npu_available
 from typing import Any, Dict, Optional
 
 from swift.utils import check_json_format, get_logger, init_process_group, is_master, set_device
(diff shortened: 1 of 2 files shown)


Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9465, opened 2026-06-01
Add lightweight TGS and MFU logging for training

WendaDeng:
# PR type
- [ ] Bug Fix
- [ * ] New Feature
- [ ] Document Updates

# PR information

HF swift sft:
- without DEVICE_TFLOPS: tgs is printed, MFU is not printed
- with DEVICE_TFLOPS=xxx: tgs and MFU are printed when state.total_flos > 0

Megatron sft:
- without DEVICE_TFLOPS: tgs is printed, MFU is not printed
- with DEVICE_TFLOPS=xxx: tgs, throughput and MFU are printed

# PR explanation:
Although `PerfMetricsLogCallback` already provides MFU logging, this change intentionally keeps the lightweight progress-log path separate. The default progress callback is responsible for writing the console progress output and `logging.jsonl`; user callbacks such as `perf_log` may run after the progress log has already been emitted, so adding MFU there does not reliably make the metric appear in the same training log entry.

This implementation only enables MFU when `DEVICE_TFLOPS` is explicitly provided, avoiding the extra device benchmark in `perf_log.py`. It also keeps the change limited to log formatting/callback code and does not patch the training step or FLOPs calculation logic.

### DIFF
--- a/swift/arguments/sft_args.py
+++ b/swift/arguments/sft_args.py
@@ -231,6 +231,8 @@ class SftArguments(SwanlabArguments, TunerArguments, BaseArguments, Seq2SeqTrain
             self.eval_strategy = 'no'
         self.training_args = TrainerFactory.get_training_args(self)
         self.training_args.remove_unused_columns = False
+        # The generic HF log patch only sees training_args, so pass the fixed-length TGS size through.
+        self.training_args.seq_length = self.packing_length or self.max_length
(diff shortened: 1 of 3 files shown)


Jintao-Huang:
👋 hi!

Thank you for your PR, but it will not be merged. I will open a new PR with similar functionality to support this feature.

PR #9468, opened 2026-06-02
Npu grpo Fix

addsubmuldiv:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

This PR fixes several NPU GRPO runtime compatibility issues with vLLM-Ascend 0.18. The added compatibility logic is scoped to the NPU / vLLM-Ascend path and should not affect GPU paths by default.

Main fixes:

- Remove the global NPU vLLM monkey patch on `torch.distributed.new_group()`. The previous patch injected `use_local_synchronization=True` for all backends and could hang vLLM-Ascend initialization / communicator creation in GRPO colocate mode.
- Add vLLM-Ascend memory API compatibility by wrapping `NPUPlatform.mem_get_info` and binding bare `torch.npu.mem_get_info()` calls to the current vLLM device. This avoids wrong-device queries and argument-binding errors during initialization and sleep/wake.
- Route non-quantized MoE routing to `torch_npu.npu_moe_init_routing_v2` when the installed vLLM-Ascend implementation still dispatches that path to `npu_moe_init_routing_custom / aclnnMoeInitRoutingCustom`.
- Add an NPU-only vLLM-Ascend MoE expert weight loader for fused `w13_weight` / `w2_weight`, so GRPO runtime weight sync can copy training-side weights into the processed expert layout expected by vLLM-Ascend.
- Handle FSDP2 Qwen MoE fused expert parameter names on the NPU path, such as `mlp.experts.gate_up_proj` / `mlp.experts.down_proj`, so they can enter the vLLM MoE loader.
- Complete weight / adapter sync initialization for vLLM server rollout mode, keeping server and colocate rollout behavior consistent for GRPO weight sync.

Known boundaries:

- vLLM-Ascend upstream has an unmerged PR that refactors `w13_weight` / `w2_weight` transpose state management for RL weight reload. Current vLLM-Ascend 0.18/main does not include that fix yet, so SWIFT keeps the NPU-only weight-loader compatibility layer.
- The FSDP2 Qwen MoE fused-name compatibility is currently a bridge for the NPU/vLLM-Ascend weight-sync path. A more general root fix would make the vLLM loader or SWIFT's generic weight-sync name mapper handle these fused MoE names natively. This PR does not claim coverage for all model families or GPU MoE paths.
- Verified coverage is limited to NPU vLLM-Ascend 0.18, Qwen3 dense / Qwen3 MoE, FSDP2 / Megatron GRPO, and server / colocate rollout.

## Experiment results

Verified on NPU:

- FSDP2 Qwen3 dense GRPO;
- FSDP2 Qwen3 MoE GRPO;
- Megatron backend GRPO;
- vLLM server rollout;
- vLLM colocate rollout;
- standalone vLLM-Ascend inference.


Jintao-Huang:
cc @hjh0119

swift/model/npu_patch/vllm_ascend_moe.py
```diff
@@ -0,0 +1,304 @@
...
+    if name.endswith(gate_up_suffix):
+        prefix = name[:-len('gate_up_proj')]
+        return [
+            f'{prefix}0.gate_proj.weight',
+            f'{prefix}0.up_proj.weight',
+        ]
+    if name.endswith(down_suffix):
+        prefix = name[:-len('down_proj')]
+        return [f'{prefix}0.down_proj.weight']
+    return None
```
hjh0119, line 154: will this qwen patch affect the weight mapping of other models?

swift/model/npu_patch/vllm_ascend_moe.py
```diff
@@ -0,0 +1,304 @@
...
+        return None
+    if name.endswith('.mlp.experts.gate_up_proj'):
+        gate_proj, up_proj = param.chunk(2, dim=1)
+        return [
+            (expanded_names[0], gate_proj.contiguous()),
+            (expanded_names[1], up_proj.contiguous()),
+        ]
+    if name.endswith('.mlp.experts.down_proj'):
+        return [(expanded_names[0], param)]
+    return None
```
hjh0119, line 172: just curious: will these be fixed in vllm_ascend in the future?

hjh0119 (approved):

hjh0119:
thanks for your contribution! I have left some minor comments

swift/model/npu_patch/vllm_ascend_moe.py
addsubmuldiv: It should not affect other models. The mapping is gated by exact Qwen-style fused expert parameter names, i.e. `.mlp.experts.gate_up_proj` and `.mlp.experts.down_proj`. For any other name, this helper returns `None` and the original weight mapping path is unchanged.

swift/model/npu_patch/vllm_ascend_moe.py
addsubmuldiv: After checking the upstream vLLM examples, I found this issue is not NPU-specific.

vLLM's official RLHF IPC FSDP EP example already shows a similar workaround on the GPU/FSDP2 path:
https://docs.vllm.ai/en/latest/examples/rl/rlhf_ipc_fsdp_ep/

In that example, HF Qwen MoE expert weights may be exposed as fused 3D tensors, while vLLM still expects checkpoint-style per-expert names during weight loading. The example therefore un-fuses these tensors before sending them to vLLM.

So this looks like a known layout mismatch on the upstream side, and we should keep tracking future vLLM/vLLM-Ascend updates. The current PR only adds a SWIFT-side temporary compatibility bridge for the NPU vLLM-Ascend runtime weight-sync path. If upstream later supports the fused Qwen MoE layout directly, this patch can be removed or disabled.

PR #9511, opened 2026-06-08
Fix transformers 5 deepspeed accelerator state

addsubmuldiv:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

This PR fixes a Transformers 5 + DeepSpeed initialization failure on NPU.

The visible error is:

```text
AttributeError: 'AcceleratorState' object has no attribute 'mixed_precision'
```

The failure happens during `Trainer` / Accelerate startup, before the first training step.

Root cause:

- `torch_npu` patches `os.getenv` and `os.environ.get` in `torch_npu.utils.patch_getenv`.
- The patch stores the original `os.environ.get` / `os.getenv` methods at import time.
- When HuggingFace Datasets runs with `num_proc >= 1`, it can replace the whole `os.environ` object while preprocessing the dataset.
- Later Accelerate writes `ACCELERATE_USE_DEEPSPEED=true` into the current `os.environ`, but the patched `os.environ.get` / `os.getenv` can still read from the old `os.environ` object.

This creates an inconsistent state where the environment variable exists in the current environment mapping but `.get()` cannot see it:

```text
"ACCELERATE_USE_DEEPSPEED" in os.environ == True
os.environ["ACCELERATE_USE_DEEPSPEED"] == "true"
os.environ.get("ACCELERATE_USE_DEEPSPEED") == None
os.getenv("ACCELERATE_USE_DEEPSPEED") == None
```

Accelerate uses `os.environ.get("ACCELERATE_USE_DEEPSPEED")` to decide whether to initialize the DeepSpeed state. Because `.get()` returns `None`, the DeepSpeed branch is skipped and `AcceleratorState.deepspeed_plugins` is not initialized, which later causes the `mixed_precision` AttributeError.

This PR fixes the NPU-specific root cause in `swift.model.npu_patch.env` by replacing the problematic `torch_npu` getenv wrapper with one that always reads from the current `os.environ` object. The previous broad `AcceleratorState` compatibility patch in `SwiftMixin` is removed.

The change is scoped to NPU environments where `torch_npu.utils.patch_getenv` is active. GPU paths are not changed.

## Experiment results

Verified on NPU:

- `transformers==5.3.0`
- `accelerate==1.13.0`
- `deepspeed==0.18.0`
- `torch==2.7.1`
- `torch_npu==2.7.1.post4`

Checks run:

```bash
python -m py_compile swift/trainers/mixin.py swift/model/npu_patch/env.py
git diff --check
```

Minimal reproduction verified:

- import `torch_npu` so its getenv patch is active
- import `swift.model` so the NPU patch is applied
- simulate HuggingFace Datasets replacing `os.environ`
- write `ACCELERATE_USE_DEEPSPEED=true`
- verify both `os.environ.get(...)` and `os.getenv(...)` return `true`

Ran an equivalent Qwen3-VL DeepSpeed ZeRO-2 SFT smoke on NPU with `--dataset_num_proc 32` and `--max_steps 1`.

Result:

- training reached `global_step=1`
- `loss=2.632`
- `grad_norm=99.34`
- final checkpoint was saved successfully


Jintao-Huang:
哈咯，这个问题是 GPU上也有这个问题嘛

addsubmuldiv:
> 哈咯，这个问题是 GPU上也有这个问题嘛

后来发现没有的，本来以为在GPU上跑过了，实际上忘记了，没有跑过。。试了一下发现GPU是能正常跑的，深入研究了一下发现NPU这边的问题主要在于torch_npu的一个patch错误地修改了os.environ.get的行为导致的，已经修改代码仅针对这一问题进行修复

Jintao-Huang (approved):

PR #9551, opened 2026-06-12
[bugfix] pin MeanMetric all_reduce tensor to float32 to avoid cross-rank dtype mismatch

### LINKED ISSUE #9550: Potential dtype inconsistency in distributed metric reduction
### Checklist / 检查清单

## Description

While training a multimodal (text + audio) model with distributed training, I encountered a hang during metric synchronization.

After adding debug logs, I found that different ranks were creating reduction tensors with different dtypes immediately before `dist.all_reduce()`, which eventually caused the collective communication to fail and block.

Debug screenshots are attached below.

Image

## Investigation

After tracing the metric pipeline, the issue appears to involve the interaction between:

- `compute_acc()` in `metrics/acc.py`
- `MeanMetric.compute()` in `metrics/utils.py`

In sequence-level accuracy mode, `compute_acc()` appends the result of `np.all(...)` directly:

```python
acc_list.append(np.all(preds[i, m] == labels[i, m]))
```

Since `np.all()` returns `numpy.bool_`, NumPy scalar types may propagate through metric accumulation and eventually affect the runtime type of `self.state` inside `MeanMetric`.

Later, `MeanMetric.compute()` constructs the reduction tensor via:

```python
tensor = torch.tensor([self.state, self.count], device=self.device)
```

Because the dtype is inferred dynamically, the resulting tensor dtype depends on the actual runtime type of `self.state`.

From my debugging logs, some ranks entered the collective operation with:

```text
torch.float32
```

while others entered with:

```text
torch.float64
```

which is consistent with the observed distributed synchronization failure.

## Expected Behavior

Metric reduction should use a consistent tensor dtype across all ranks, regardless of whether intermediate metric values originate from Python scalars or NumPy scalars.

## Additional Information

Attached screenshots show:

1. The dtype mismatch observed across ranks before `dist.all_reduce()`.
2. The intermediate metric values involved in the dtype divergence.

I have investigated the issue locally and prepared a potential fix together with a small regression test. If the analysis above looks reasonable, I'd be happy to open a PR for discussion.

### How to Reproduce / 如何复现

The issue was observed during distributed training of a multimodal (text + audio) model using multiple GPUs.

1. Launch distributed training/evaluation with sequence-level accuracy metrics enabled.
2. Run evaluation on a multimodal dataset including text and audio.
3. Add debug logs before `MeanMetric.compute()` performs `dist.all_reduce()`.
4. Observe that some ranks construct reduction tensors with `torch.float32` while others construct `torch.float64`.
5. The subsequent `dist.all_reduce()` may hang or fail due to the dtype mismatch across ranks.

### Minimal Reproducer

```python
import numpy as np
import torch

state = 0.

acc_list = [
    np.bool_(True),
    np.bool_(False),
]

state += sum(acc_list)

print(type(state))
# <class 'numpy.float64'>

tensor = torch.tensor([state, 1])
print(tensor.dtype)
# torch.float64
```

In contrast, using Python `bool` values produces a different inferred dtype:

```python
state = 0.

acc_list = [
    True,
    False,
]

state += sum(acc_list)

print(type(state))
# <class 'float'>

tensor = torch.tensor([state, 1])
print(tensor.dtype)
# torch.float32
```

I have investigated the issue locally and prepared a potential fix. If the analysis above looks reasonable, I'd be happy to open a PR for discussion.

HaozheZhang6:
# PR type
- [x] Bug Fix

# PR information

Fixes #9550.

`MeanMetric.compute()` (`swift/metrics/utils.py`) builds the all-reduce tensor with an **inferred** dtype:

```python
tensor = torch.tensor([self.state, self.count], device=self.device)
```

In sequence-level accuracy, `compute_acc()` appends `np.all(...)` (a `numpy.bool_`). Accumulating that into `self.state` promotes it from a python `float` to `numpy.float64` on the ranks that take that path. `torch.tensor([...])` then infers **float64** on those ranks but **float32** on ranks where `self.state` stayed a python float. The ranks enter `dist.all_reduce()` with mismatched dtypes and the collective hangs (matching the debug screenshots in #9550).

Pinning the tensor to `float64` makes every rank reduce with the same dtype regardless of `self.state`'s runtime type. float64 also keeps `count` exact for large totals (float32 loses integer precision above 2^24).

## Experiment results

Reproducing the hang needs a multi-rank run; minimal demo of the dtype inference (one rank's state promoted to `numpy.float64` via `np.bool_`, the other left a python float):

```
rank-A state type: float | rank-B state type: float64
CURRENT  -> rankA=torch.float32  rankB=torch.float64  all_reduce MATCH=False
FIX      -> rankA=torch.float64  rankB=torch.float64  all_reduce MATCH=True
```

### DIFF
--- a/swift/metrics/utils.py
+++ b/swift/metrics/utils.py
@@ -101,7 +101,7 @@ class MeanMetric(Metric):
 
     def compute(self):
         if dist.is_initialized():
-            tensor = torch.tensor([self.state, self.count], device=self.device)
+            tensor = torch.tensor([self.state, self.count], dtype=torch.float64, device=self.device)
             dist.all_reduce(tensor, op=dist.ReduceOp.SUM, group=self.group)
             self.state, self.count = tensor[0].item(), int(tensor[1].item())
         if self.count == 0:
--- a/swift/metrics/utils.py
+++ b/swift/metrics/utils.py
@@ -101,7 +101,7 @@ class MeanMetric(Metric):
 
     def compute(self):
         if dist.is_initialized():
-            tensor = torch.tensor([self.state, self.count], dtype=torch.float64, device=self.device)
+            tensor = torch.tensor([self.state, self.count], dtype=torch.float32, device=self.device)
             dist.all_reduce(tensor, op=dist.ReduceOp.SUM, group=self.group)
             self.state, self.count = tensor[0].item(), int(tensor[1].item())
         if self.count == 0:


Jintao-Huang:
thanks!

Jintao-Huang:
Let's use float32 for now — I'm concerned that some hardware may not support float64.

Jintao-Huang (approved):

HaozheZhang6:
Makes sense, switched to float32.

PR #9562, opened 2026-06-15
Update gkd_loss.py teacher_prompt supports two formats:

yhy19:
teacher_prompt supports two formats:
        1. str: Used only as the user content for the teacher, replacing the last user message.
        2. list[dict]: A complete list of messages (including system/user/assistant),
           which directly replaces the teacher's messages (retaining the assistant from the original data as the response).

# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

## Experiment results


hjh0119:
If replacing the full messages, would using the `teacher_messages` key be more appropriate?

yhy19:
> If replacing the full messages, would using the `teacher_messages` key be more appropriate?

That makes perfect sense, I'll update it right away.

PR #9598, opened 2026-06-18
feat(packing): add opt-in packing_strategy (binpack | sequential) for order-preserving packing

yuchenwang3:
## What

The packing path uses `binpacking.to_constant_volume` (best-fit-decreasing), which **reorders** samples (by length) and changes which samples land in each pack. For recipes that rely on a **sequential sampler** — where the sample order and pack boundaries are themselves part of the recipe (e.g. domain weighting encoded in the sample order, as in StepFun Step-3.5) — this reordering breaks faithful reproduction.

This adds an opt-in `packing_strategy`:
- **`binpack`** (default, unchanged): best-fit-decreasing bin packing.
- **`sequential`**: order-preserving greedy packing (next-fit) — keep a single open pack, flush it when the next sample doesn't fit, and carry the trailing unfinished pack to the next batch. Sample order and pack boundaries follow the input order (use `packing_num_proc=1` for a single global ordering).

## Change

- `calculate_matched_group(..., strategy='binpack')`: add the `sequential` (next-fit) branch; the default `binpack` path is unchanged.
- `PackingDataset` / `IterablePackingDataset`: accept `packing_strategy` and pass it through to `calculate_matched_group`.
- `BaseArguments.packing_strategy: Literal['binpack', 'sequential'] = 'binpack'`, threaded through `sft.py`.
- zh/en command-line docs.

**Default behavior is unchanged** (`binpack`).

## Testing

Verified syntactically (`py_compile`) and traced the call path (args → `sft.py` → `PackingDataset`/`IterablePackingDataset` → `calculate_matched_group`). I could not run the full training suite locally; relying on CI. The `sequential` algorithm (order-preserving next-fit with carry-over) was used in real Qwen3.5-35B SFT to reproduce a sequential-sampler recipe.

## Note

Named `sequential` (not `first_fit`) because the algorithm keeps a single open pack and flushes on overflow — i.e. next-fit — and because the user-facing property is order preservation. Happy to rename if maintainers prefer.


yuchenwang3:
@Jintao-Huang gentle ping 🙏 Opt-in `packing_strategy` (binpack|sequential), default unchanged / fully backward-compatible. Could you take a look when you have a moment? Happy to address any feedback.

Jintao-Huang (approved):

Jintao-Huang:
thanks!

Issue #6387, opened 2025-10-31
Qwen3-Next-80B-A3B-Instruct 8卡H20，lora微调报错OOM   Failed to CUDA calloc async 136 bytes


xiaotianns:
**Describe the bug**
Qwen3-Next-80B-A3B-Instruct 8卡H20，lora微调报错OOM
运行命令
`PYTORCH_CUDA_ALLOC_CONF='expandable_segments:True' \
NPROC_PER_NODE=8 \
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 \
megatron sft \
    --load /model/Qwen3-Next-80B-A3B-Instruct-mcore \
    --dataset 'high_quality_data_0101_1022_265w_converted.json' \
    --train_type lora \
    --lora_rank 8 \
    --lora_alpha 32 \
    --target_modules all-linear \
    --expert_model_parallel_size 2 \
    --moe_permute_fusion true \
    --moe_grouped_gemm true \
    --moe_shared_expert_overlap true \
    --moe_aux_loss_coeff 1e-3 \
    --micro_batch_size 1 \
    --global_batch_size 8 \
    --recompute_granularity full \
    --recompute_method uniform \
    --recompute_num_layers 1 \
    --max_epochs 3 \
    --finetune true \
    --cross_entropy_loss_fusion true \
    --lr 1e-4 \
    --lr_warmup_fraction 0.05 \
    --min_lr 1e-5 \
    --save /megatron_output/Qwen3-Next-80B-A3B-Instruct \
    --save_interval 82976 \
    --max_length 512 \
    --num_workers 8 \
    --dataset_num_proc 8 \
    --no_save_optim true \
    --no_save_rng true \
    --sequence_parallel true \
    --attention_backend flash \
    --model_author swift \
    --model_name swift-robot`

报错：
`[INFO:swift] [rank0] model_parameter_info: PeftModelForCausalLM: 41583.6741M Params (563.9885M Trainable [1.3563%]), 0.0000M Buffers.
 > number of parameters on (tensor, pipeline) model parallel rank (0, 0): 41583674112
sharded_state_dict metadata loaded from the checkpoint: {'distrib_optim_sharding_type': 'fully_sharded_model_space'}
 loading distributed checkpoint from /ssd4/nietianyu/workspace/ms-swift/model/Qwen3-Next-80B-A3B-Instruct-mcore at iteration 1
[...]

PR #9602, opened 2026-06-18
feat(megatron): add nccl_comm_warmup to avoid iteration-1 NCCL cudaMalloc OOM (#6387)

yuchenwang3:
## What

Adds an opt-in `nccl_comm_warmup` flag that eagerly creates the NCCL communicators before the training loop, to avoid an iteration-1 OOM.

## Why

NCCL communicators are created lazily on first use. For the per-step collectives that ms-swift/Megatron issue inside the training step (e.g. the dp/cp loss all-reduce and grad-sync coalescing), that first use happens at the **iteration-1 memory peak**. At that point NCCL's internal `cudaMalloc` for the communicator buffers can fail with:

```
Failed to CUDA calloc async N bytes
```

so training dies on the very first step on large models / tight-memory configs. This is the error reported in #6387 (Qwen3-Next-80B-A3B, 8×H20), which was only worked around by changing the parallel layout (`--expert_model_parallel_size 8`) rather than fixed in code. We hit the same failure on Qwen3.5-35B-A3B.

## Change

When `--nccl_comm_warmup true`, right before `on_train_begin` (in `setup_model_training`), fire one numerically-inert 1-element `all_reduce` on each parallel group (dp, dp+cp, cp, tp, pp, model, embedding, position-embedding). This forces the communicators to be allocated while memory is still free; groups that don't exist for the current parallel config are skipped. Default is `False`, so behavior is unchanged unless explicitly enabled.

## Notes

- Opt-in and default-off; zero effect on existing runs.
- The dummy all-reduce is a no-op numerically (a 1-element zero tensor).
- Verified the two files compile; could not run a multi-GPU job locally — relying on CI / users who hit #6387.

### DIFF
--- a/swift/megatron/arguments/megatron_args.py
+++ b/swift/megatron/arguments/megatron_args.py
@@ -546,6 +546,9 @@ class MegatronArguments(RLHFMegatronArgumentsMixin, MegatronTunerMixin):
     overlap_param_gather: bool = False
     overlap_param_gather_with_optimizer_step: bool = False
     align_grad_reduce: bool = True
+    # Eagerly create NCCL communicators before the training loop to avoid the lazy
+    # first-use allocation hitting the iteration-1 memory peak (Failed to CUDA calloc async).
+    nccl_comm_warmup: bool = False
     virtual_pipeline_model_parallel_size: Optional[int] = None
     microbatch_group_size_per_vp_stage: Optional[int] = None
     pipeline_model_parallel_layout: Optional[str] = None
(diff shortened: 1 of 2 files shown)


yuchenwang3:
@Jintao-Huang gentle ping 🙏 Small self-contained `nccl_comm_warmup` flag to avoid the first-iteration NCCL hang, default off. Would appreciate a quick look when you have time.

Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9606, opened 2026-06-20
[bugfix] guard load_audio against bytes input

he-yufeng:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

`load_audio()` accepts `Union[str, bytes]`. When the first `librosa.load()` fails (e.g. an audio format soundfile can't read on its own), it falls into the `except` block and checks `audio.startswith(('http://', 'https://'))`. For `bytes` input this raises `TypeError: a bytes-like object is required, not 'str'`, masking the real decode error — and `bytes` is a documented input type (the dataset can hand the template raw audio bytes, and `load_file` already handles `bytes`).

This guards the scheme check with `isinstance(audio, str)`, so byte audio takes the non-URL fallback path instead of crashing on the type mismatch. String/URL behavior is unchanged.

## Experiment results

Added `test_load_audio_bytes_input_does_not_crash_on_fallback` in `tests/general/test_template.py`: it injects a stub `librosa` whose first `load` raises, passes `bytes` audio, and asserts `load_audio` reaches the fallback without a `TypeError`. The test fails on the previous code (`TypeError` at the `startswith` line) and passes with this change. flake8 / yapf / isort are clean on both files.

```
$ python -m pytest tests/general/test_template.py::test_load_audio_bytes_input_does_not_crash_on_fallback -q
1 passed
```

### DIFF
--- a/swift/template/vision_utils.py
+++ b/swift/template/vision_utils.py
@@ -103,6 +103,10 @@ _T = TypeVar('_T')
 
 def _check_path(path: str) -> Union[str, None]:
     """If it is a path, return the string; if it is base64, return None."""
+    if not isinstance(path, str):
+        # bytes audio/image data is not a path; let the caller fall back to it
+        # instead of crashing on the str-only checks below (e.g. startswith).
+        return None
     MAX_PATH_HEURISTIC = 2000
     if len(path) > MAX_PATH_HEURISTIC:
         return
--- a/tests/general/test_template.py
+++ b/tests/general/test_template.py
@@ -137,6 +137,35 @@ def test_load_video_minicpmv_handles_zero_fps():
     assert len(frames) == 4
 
 
+def test_load_audio_bytes_input_does_not_crash_on_fallback(monkeypatch):
+    import sys
+    import types
+
+    from swift.template import vision_utils
+
+    calls = []
+
+    fake_librosa = types.ModuleType('librosa')
+
+    def fake_load(audio_io, sr):
+        calls.append(audio_io)
+        if len(calls) == 1:
+            # First attempt fails (e.g. a format soundfile can't read), forcing
+            # the except branch that used to call bytes.startswith and crash.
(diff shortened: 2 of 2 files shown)


he-yufeng:
Extended this to fully guard `load_audio` against bytes input. The first commit fixed the `audio.startswith(...)` check (line 305); but on the same except-branch fallback, `_check_path(audio)` was still called with bytes and crashed on its str-only checks (`data.startswith('data:')`). The second commit makes `_check_path` return `None` for non-str input so the caller falls back to the raw bytes. The existing fallback test now exercises the real `_check_path` (the mock is removed) and a focused `_check_path` bytes test is added.

he-yufeng:
Following up here — load_audio currently assumes a path and breaks on raw bytes input; this guards that case. It's been open a few weeks now — could a maintainer take a look? Thanks.

he-yufeng:
Second nudge, a month on. Still mergeable and green. The gap remains: `load_audio` assumes a filesystem path, so a caller holding raw bytes has no way through. Happy to rework or close if the direction isn't wanted.

Excelius-Wang (changes requested):
The new fallback regression fails at the current head:

```text
TypeError: fake_load() got an unexpected keyword argument 'mono'
```

`_load_audio_librosa()` passes `mono=mono` to both decoder calls, but the test double only accepts `audio_io` and `sr`, so it never reaches the intended simulated decode failure.

After giving the test double a matching signature, I checked the same scenario on both this PR and its base: it passes on both, and `_check_path` is never called. The current bytes fallback already bypasses it via `if isinstance(audio, str) else audio`.

Could you update the test signature and clarify which current production path still needs the new guard? The direct `_check_path(bytes)` test covers the helper, but the load_audio regression no longer demonstrates the failure described in the PR. These were offline CPU checks with a simulated decoder failure, not codec validation.

PR #9608, opened 2026-06-20
[bugfix] guard _repair_ms_bench against an empty messages list

he-yufeng:
## Summary

`_repair_ms_bench` (the `repair_messages` hook for the `iic/ms_bench` dataset) reads `messages[0]` right after `ast.literal_eval`, so a row whose messages are empty — e.g. the literal string `"[]"` — raises `IndexError: list index out of range` and aborts the whole ms_bench dataset load instead of just skipping that one row.

The function already returns `None` to skip rows it can't use (the MOSS case), and `MessagesPreprocessor` drops `None` rows, so returning `None` for an empty list is the same, consistent behaviour.

## Fix

An early `if not messages: return None` guard right after the parse.

## Test

Added `tests/general/test_repair_ms_bench.py` (pure unit tests, no network): empty `"[]"` / `[]` returns `None`, the default system message is stripped, a normal conversation passes through, and a MOSS row is skipped. The empty case raises `IndexError` without the fix.

Verified locally: `pytest tests/general/test_repair_ms_bench.py` (4 passed); `flake8` / `isort` / `yapf` clean.

### DIFF
--- a/swift/dataset/dataset/llm.py
+++ b/swift/dataset/dataset/llm.py
@@ -93,6 +93,10 @@ register_dataset(
 def _repair_ms_bench(messages: str) -> Optional[List[Dict[str, str]]]:
     if isinstance(messages, str):
         messages = ast.literal_eval(messages)
+    if not messages:
+        # A row with no messages can't be repaired; skip it like the MOSS case
+        # below instead of crashing the whole dataset load on messages[0].
+        return None
     default_system = 'You are a helpful assistant.'
     messages: List[Dict[str, str]]
     if messages[0]['from'] == 'system' and messages[0]['value'] == default_system:
--- a/tests/general/test_repair_ms_bench.py
+++ b/tests/general/test_repair_ms_bench.py
@@ -1,0 +1,46 @@
+import unittest
+
+from swift.dataset.dataset.llm import _repair_ms_bench
+
+
+class TestRepairMsBench(unittest.TestCase):
+    """Pure unit tests for the ms_bench messages repair function (no network)."""
+
+    def test_empty_messages_returns_none(self):
+        # An empty row can't be repaired; it must be skipped (None) like the MOSS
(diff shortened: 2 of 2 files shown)


he-yufeng:
Friendly nudge on this small guard — without it `_repair_ms_bench` blows up on the malformed-entry case instead of skipping it. It's been open a few weeks and stays self-contained; happy to rebase onto current main or fold in anything you'd want changed.

he-yufeng:
Second nudge, a month after the first. The guard is still one line and still mergeable with green checks: without it `_repair_ms_bench` dies on the first non-directory entry the glob returns. Happy to close it if you'd rather handle it differently.

hjh0119:
Could you provide more context on where this was observed? I tried the current `iic/ms_bench` dataset and couldn’t find any empty `messages` rows

he-yufeng:
Fair question. I just scanned the first 38861 rows of the current ms_agent_bench_v1_sft_sample.jsonl myself and found zero rows with an empty messages list, same as your check. So the current published dataset does not exhibit the case.

The guard's basis is code-level: _repair_ms_bench indexes messages[0] immediately after ast.literal_eval, so a single row whose messages parse to [] aborts the whole dataset load with IndexError instead of being skipped the way the MOSS branch skips unusable rows. It is one line and costs nothing when the data is clean.

If you'd rather not carry a guard for a case the data doesn't currently show, I'm fine closing this.

he-yufeng:
@hjh0119 Fair question — this came from reading the code path, not from hitting a bad row on the real dataset, so I can't point at an actual `"[]"` row in the current iic/ms_bench snapshot. The guard is the same shape as the existing MOSS `None`-skip two lines below: any future row whose messages literal evaluates to an empty list crashes the whole load with IndexError instead of being skipped. If you'd rather not guard a hypothetical row, I'm fine closing it; the check is two lines and easy to revive if the dataset ever produces one.

PR #9611, opened 2026-06-22
fix(dataset): propagate system message to rejected_messages in DPO preprocessing

he-yufeng:
When a DPO dataset row has both a `system` key and `rejected_messages`, the system message is never added to `rejected_messages`. The recursive call to `preprocess` only passes `{'messages': row['rejected_messages']}` — the `system` key is dropped on the way in.

DPO training requires chosen and rejected to have the same prompt context; this silently produces asymmetric encoding. Anything downstream that encodes both sides will see different conversation prefixes, which corrupts the training signal.

One-line fix: pass `'system': row.get('system')` to the recursive call so the system message is forwarded.

Also adds a regression test covering the scenario.

### DIFF
--- a/swift/dataset/preprocessor/core.py
+++ b/swift/dataset/preprocessor/core.py
@@ -514,7 +514,7 @@ class MessagesPreprocessor(RowPreprocessor):
     def preprocess(self, row: Dict[str, Any]) -> Optional[Dict[str, Any]]:
         if 'rejected_messages' in row:
             row['rejected_messages'] = MessagesPreprocessor.preprocess(
-                self, {'messages': row['rejected_messages']})['messages']
+                self, {'messages': row['rejected_messages'], 'system': row.get('system')})['messages']
         messages = row['messages']
         if self.inner_key is not None:
             messages = messages[self.inner_key]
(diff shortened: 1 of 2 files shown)


Jintao-Huang:
The system content is usually placed inside `messages` and `rejected_messages`.

he-yufeng:
You're right that in the common format the system turn is already embedded as the first message in both `messages` and `rejected_messages`, and there this change is a no-op.

The case it fixes is when the dataset carries `system` as a separate top-level column (the `system` / `system_prompt` mapping). For those rows `preprocess` pops `system` and prepends it to `messages` via `to_std_messages` / `sharegpt_to_messages`, but the recursive call for `rejected_messages` was built as `{'messages': row['rejected_messages']}` without `system`, so the system turn only lands on the chosen side. DPO then trains on a pair whose chosen and rejected halves have different prefixes.

The regression test builds exactly such a row: with the fix, system is first in both lists; reverting the one-line change makes it fail with `'user' != 'system'` on `rejected_messages[0]`.

Jintao-Huang:
Thanks, this PR will not be merged.

PR #9612, opened 2026-06-22
fix(template): preserve non-string prefix elements in _replace_system

he-yufeng:
`_replace_system` is responsible for building the no-system variant of `prefix` by replacing the `{{SYSTEM}}` placeholder in string elements. The current implementation uses a filter comprehension:

```python
# before
return [p.replace('{{SYSTEM}}', '') for p in prefix if isinstance(p, str)]
```

The `if isinstance(p, str)` clause silently **drops** every non-string element. Token-ID lists like `['bos_token_id']` are legitimate prefix components — templates such as `ziya`, `bluelm`, and `emu3_chat` (and others in `baai.py`) use `prefix=[['bos_token_id'], '{{SYSTEM}}']`. When `_replace_system` runs (i.e. no system message is provided), the BOS token-ID entry disappears from `meta.prefix`, so every training and inference sequence for these templates starts without the expected leading token.

**Fix:** change the filter to a conditional expression so non-strings pass through unchanged:

```python
# after
return [p.replace('{{SYSTEM}}', '') if isinstance(p, str) else p for p in prefix]
```

Adds a focused unit test in `tests/general/test_template_meta.py` that instantiates a `TemplateMeta` with a mixed prefix and asserts the list element survives `__post_init__`.

### DIFF
--- a/swift/template/template_meta.py
+++ b/swift/template/template_meta.py
@@ -69,7 +69,7 @@ class TemplateMeta:
 
     @staticmethod
     def _replace_system(prefix: Prompt) -> Prompt:
-        return [p.replace('{{SYSTEM}}', '') for p in prefix if isinstance(p, str)]
+        return [p.replace('{{SYSTEM}}', '') if isinstance(p, str) else p for p in prefix]
 
     def _check_template_meta(self):
         # check
--- a/tests/general/test_template_meta.py
+++ b/tests/general/test_template_meta.py
@@ -1,0 +1,24 @@
+# Copyright (c) ModelScope Contributors. All rights reserved.
+from swift.template import TemplateMeta
+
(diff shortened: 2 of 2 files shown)


Jintao-Huang:
thanks!

Jintao-Huang:
Please pass the lint checks:

```shell
pip install pre-commit
pre-commit run --all-files
```

he-yufeng:
Done. Pushed the yapf fix in 3ec09580 (it was just formatting on the new regression test). `pre-commit run` comes back clean on both touched files now, so this should be ready for another look.

Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9639, opened 2026-06-24
Fix: MiniCPM-V 4.6 training hangs on text-only samples with DeepSpeed

randydl:
# Fix: MiniCPM-V 4.6 training hangs on text-only samples with DeepSpeed

## Problem

When training MiniCPM-V 4.6 with DeepSpeed ZeRO, the training process hangs if the dataset contains pure-text samples (no image/video). This is because:

- For text-only samples, `MiniCPMV4_6Model.forward()` skips the vision encoder (`vision_tower` + `merger`) entirely since `pixel_values` and `pixel_values_videos` are both `None`.
- Under DeepSpeed ZeRO, parameters are sharded across GPUs and only gathered on-demand via all-gather when a computation touches them.
- When one GPU processes a text-only sample (no vision compute) while another processes an image sample (needs vision parameters), the all-gather synchronization deadlocks — one side never triggers the gather that the other side is waiting for.

## Solution

Add a `_post_encode` method to `MiniCPMV4_6Template` that detects text-only samples under DeepSpeed and runs a minimal dummy image through the full vision pipeline (`vision_tower` → `merger`). The dummy features are then zeroed out via `image_embeds.mean() * 0.` and added to the text embeddings, which:

- Forces DeepSpeed to all-gather all vision model parameters, preventing the deadlock
- Is mathematically a no-op (adds zero), so it does not affect training results

The dummy image uses the smallest valid patch grid (`target_sizes=[[4, 4]]`), which works for both 16x and 4x downsample modes, producing only 1 visual token in 16x mode — negligible compute overhead.

## Changes

- `swift/template/templates/minicpm.py`: Add `_post_encode` to `MiniCPMV4_6Template`, add `is_deepspeed_enabled` import

### DIFF
--- a/swift/template/templates/minicpm.py
+++ b/swift/template/templates/minicpm.py
@@ -9,7 +9,7 @@ from packaging import version
 from torch import nn
 from typing import Any, Dict, List, Literal, Optional
 
-from swift.utils import get_env_args, get_packed_seq_params
+from swift.utils import get_env_args, get_packed_seq_params, is_deepspeed_enabled
 from ..base import Template
 from ..constant import LLMTemplateType, MLLMTemplateType
 from ..register import TemplateMeta, register_template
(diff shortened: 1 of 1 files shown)


Jintao-Huang:
Hi, please merge the main branch.

Then pass the lint tests:
```
pip install pre-commit
pre-commit run --all-files
```

randydl:
> Hi, please merge the main branch.
> 
> Then pass the lint tests:
> 
> ```
> pip install pre-commit
> pre-commit run --all-files
> ```

Fixed all tests

Jintao-Huang (approved):

Jintao-Huang:
I tested it and ZeRO-3 still has cases where it gets stuck.

randydl:
> I tested it and ZeRO-3 still has cases where it gets stuck.

This fix resolves hanging for ZeRO-2, yet ZeRO-3 still hangs. I haven’t figured out the underlying cause so far, please help review this issue.

Issue #9545, opened 2026-06-12
swanlab crashes on non-main GRPO ranks


rxqy:
### Checklist / 检查清单

- [x] I have searched existing issues, and this is a new bug report. / 我已经搜索过现有的 issues，确认这是一个新的 bug report。

### Bug Description / Bug 描述

When running GRPO with distributed training and `--report_to swanlab`, non-main ranks can crash during rollout/vLLM profiling with:

```text
RuntimeError: No active Run. Call swanlab.init() first.
```

The crash occurs during rollout, around _move_model_to_vllm():
```text
File ".../swift/rlhf_trainers/rollout_mixin.py", line 943, in _fast_infer
    self._move_model_to_vllm()
File ".../swift/rlhf_trainers/utils.py", line 626, in profiling_context
    if 'swanlab' in trainer.args.report_to and swanlab.get_run() is not None and is_main_process:
RuntimeError: No active Run. Call swanlab.init() first.
```

### How to Reproduce / 如何复现

My training script:

```text
swift rlhf \
    --rlhf_type grpo \
    --model $MODEL \
    --use_vllm true \
    --vllm_mode colocate \
    --vllm_gpu_memory_utilization 0.4 \
    --vllm_tensor_parallel_size 8 \
    --vllm_max_model_len 20480 \
    --beta 0. \
    --num_generations $NUM_GENERATIONS \
    --external_plugins reward_plugin.py \
    --reward_funcs some_random_reward \
    --output_dir $OUTPUT_DIR \
    --tuner_type full \
    --dataset some_random_jsonl.files \
    --dataloader_drop_last true \
    --dataloader_persistent_workers true \
    --dataloader_num_workers 8 \
    --load_from_cache_file false \
    --warmup_ratio 0.05 \
    --num_train_epochs 4 \
    --per_device_train_batch_size 1 \
    --gradient_checkpointing true \
    --gradient_accumulation_steps 8 \
    --learning_rate 5e-7 \
    --max_length 18000 \
    --max_completion_length 2048 \
    --save_steps 100 \
    --save_total_limit 999 \
    --save_only_model true \
    --logging_steps 5 \
    --report_to swanlab tensorboard \
    --eval_strategy no \
    --eval_steps 100 \
    --deepspeed zero3
```

### Additional Information / 补充信息

swift version: 
[INFO:swift] Start time of running main: 2026-06-12 15:54:39.049291
[INFO:swift] swift.__version__: 4.4.0.dev0

[...]

PR #9666, opened 2026-06-30
Adjust the order to avoid errors

1787648106:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Adjusted the order of conditions within the `profiling_context` in `rlhf_trainers/utils.py` to prevent errors during multi-process training.

## Experiment results

### DIFF
--- a/swift/rlhf_trainers/utils.py
+++ b/swift/rlhf_trainers/utils.py
@@ -579,10 +579,10 @@ def profiling_context(trainer, name: str):
     elif hasattr(trainer, 'is_main_process'):
         is_main_process = trainer.is_main_process
 
-    if 'wandb' in trainer.args.report_to and wandb.run is not None and is_main_process:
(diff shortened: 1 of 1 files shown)


1787648106:
@tastelikefeet  @Jintao-Huang Could you please take a look at this PR? Thanks!

hjh0119:
Could you describe the issue you ran into?

1787648106:
> Could you describe the issue you ran into?

In multi-process training, the active run typically resides in the main process. Since the condition in `utils.py` is evaluated from left to right, a "No active run.  Call swanlab.init() first." error is raised for any process other than the main process.

hjh0119:
The issue you met has already fixed in https://github.com/modelscope/ms-swift/pull/9547

same issue https://github.com/modelscope/ms-swift/issues/9545

1787648106:
> #9545

ok, thanks

PR #9710, opened 2026-07-09
fix: flush asynchronous JSONL logs on training end

Alwin4Zhang:
## What does this PR do?

- Add `flush()` and `close()` lifecycle methods to `JsonlWriter`.
- Wait for pending asynchronous writes before shutting down the writer thread.
- Propagate asynchronous write exceptions back to the caller.
- Close the Megatron logging writer when training ends.
- Add a regression test covering asynchronous writer shutdown.

## Motivation

Megatron writes `logging.jsonl` through a daemon thread. The writer was not flushed or closed when training finished, so the process could exit before pending log entries were written, resulting in incomplete logging files.

## Tests

- Added `test_async_jsonl_close` to verify that all queued records are persisted before `close()` returns.
- Verified that appending after the writer is closed raises `RuntimeError`.
- Passed Python compilation and `git diff --check`.

### DIFF
--- a/swift/megatron/callbacks/print.py
+++ b/swift/megatron/callbacks/print.py
@@ -32,6 +32,7 @@ class PrintCallback(MegatronCallback):
         self.jsonl_writer = JsonlWriter(logging_path, enable_async=True, write_on_rank='last')
 
     def on_train_end(self):
+        self.jsonl_writer.close()
         self.training_bar.close()
         self.training_bar = None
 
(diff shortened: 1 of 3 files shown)


Excelius-Wang (approved):
Calling close on normal training shutdown gives the daemon writer a chance to drain its queue before the process exits.

I ran the new persistence test and added checks for flush followed by more writes, repeated close, and closing an unused writer. I also simulated a background write failure: flush and close report the error, and the worker exits without leaving join blocked. All six IO tests pass. I checked the Megatron callback wiring in source, but didn't run Megatron training or distributed logging.

PR #9741, opened 2026-07-13
fix(security): block SSRF via multimodal media URLs in load_file

### LINKED ISSUE #9740: Unauthenticated SSRF via multimodal image_url/audio/video in swift deploy (no URL filtering)
### Checklist / 检查清单

### Bug Description / Bug 描述

reported on 12 June 2026 https://github.com/modelscope/ms-swift/security/advisories/GHSA-7mpg-jvqj-ccxr

### Summary

The ms-swift deployment server (`swift deploy`, OpenAI-compatible) fetches multimodal media URLs (`image_url`, `audio_url`, `video_url`) server-side with no validation of the target, and follows redirects. The server binds `0.0.0.0` and runs without authentication by default. An unauthenticated client of a multimodal deployment can therefore make the server issue requests to internal-only services and cloud instance-metadata endpoints. Confirmed against the `load_file` / `load_image`: the functions fetched an internal URL with no filter and followed a redirect.

### Details

`swift/template/vision_utils.py` `load_file` (~line 141) fetches any `http(s)` path with no checks:

```python
if path.startswith('http'):
    response = session.get(path, **request_kwargs)   # no scheme/IP filter; follows redirects
```

There is no loopback/private/link-local/metadata filtering anywhere in the function (no `ipaddress`, `is_global`, `127.0.0.1`, `localhost`, `169.254`, or `allow_redirects` handling). Reached via `/v1/chat/completions`: a message `image_url.url` is extracted in `swift/template/template_inputs.py` `remove_messages_media` (~line 127) into the images list, then `swift/template/base.py` `load_image` (~line 274) calls `load_file`. Defaults (`swift/arguments/deploy_args.py`): `host='0.0.0.0'` (~line 36), `api_key=None` (~line 38); `swift/pipelines/infer/deploy.py` `_check_api_key` (~line 116) passes when `api_key is None`. So the endpoint is unauthenticated and network-exposed by default.

### PoC

scripts/poc_msswift_ssrf.md.

```
POST /v1/chat/completions
{"model":"<vl-model>","messages":[{"role":"user","content":[
  {"type":"image_url","image_url":{"url":"http://169.254.169.254/latest/meta-data/iam/security-credentials/"}}]}]}
```

Validated against the `load_file` / `load_image`: `load_file("http://127.0.0.1:8456/secret.png")` hit the local internal listener and returned the bytes (PNG magic confirmed); `load_image` decoded it; `load_file("http://127.0.0.1:8456/redir")` followed a 302 to `/secret.png` (so an attacker host can 302 to `http://169.254.169.254/...`). Source confirms no URL filter, default host `0.0.0.0`, and `api_key=None` (no auth).

### Impact

An unauthenticated client of a multimodal ms-swift deployment can coerce server-side requests to internal services and cloud instance-metadata endpoints (credential theft), and use redirects to reach metadata even if a host allowlist is later added. This is unauthenticated SSRF on the default configuration.

_No response_

he-yufeng:
Fixes #9740 (GHSA-7mpg-jvqj-ccxr).

## The bug

`swift deploy` runs the OpenAI-compatible server binding `0.0.0.0` with `api_key=None` by default, so it is unauthenticated and network-exposed. A multimodal chat request carries `image_url` / `audio_url` / `video_url`, and every one of them funnels into `swift/template/vision_utils.py::load_file` (`load_image` → `load_file`; `_load_audio_librosa` / `_load_audio_soundfile_pyav` → `load_file`; `_resolve_video_local_path` → `load_file`). `load_file` fetched any `http(s)` URL server-side with no target validation and followed redirects.

So an unauthenticated client can point a media URL at loopback, a private range, or the cloud instance-metadata endpoint (`http://169.254.169.254/...`) and read the response back. That is SSRF, and the reported 302 redirect bypass survives a naive host allowlist.

## The fix

`_assert_media_url_allowed(url)` resolves the host and rejects it when any resolved address is private / loopback / link-local / reserved / multicast / unspecified, and rejects non-`http(s)` schemes. `load_file` now follows redirects manually and re-validates every hop, so a public URL cannot bounce into an internal address.

It is secure by default. A trusted deployment that legitimately serves media from an internal host can opt out with `SWIFT_ALLOW_INTERNAL_MEDIA_URLS=1`.

## Tests

`tests/general/test_vision_utils_ssrf.py` covers the blocked ranges (metadata / loopback / private / unspecified / non-http scheme), the public-allowed case, and the opt-out.

## Limitation

This validates each hop's resolved address, which stops the reported PoCs (direct metadata/loopback/private and the 302 bypass). It does not fully close a DNS-rebinding race between the resolution check and the socket connect; pinning the validated IP into the connection would be a reasonable follow-up, happy to add it here if you prefer.


he-yufeng:
Thanks for the careful review. Pushed `ebee680b` addressing all three points.

**1. Audio fallback bypass (critical) — fixed.** This was a real hole. `_load_audio_librosa` wraps `load_file` in `except Exception`, so the `ValueError` the guard raises for an internal URL got swallowed and the `audioread`/ffmpeg fallback then re-fetched the same URL unvalidated. I now validate the URL up front, before the `try`, so an internal `audio_url` is rejected before either path can run. Added a regression test that stubs `librosa`/`audioread` and asserts the ffmpeg fallback is never reached for a loopback URL (it fails against the previous commit). `_load_audio_soundfile_pyav` is not affected: its soundfile/pyav fallbacks operate on the `BytesIO` that `load_file` already returned, so they never re-fetch the URL.

**2. DNS rebinding (high) — partly addressed, the rest scoped as a follow-up.** You're right that validating the hostname and then letting the client re-resolve leaves a rebinding window. What this PR does cover: it re-validates on every redirect hop (so redirect-based pivots are closed) and it blocks every static internal target, which is the unauthenticated `http://169.254.169.254/...` / `http://127.0.0.1/...` case from the issue PoC. Fully closing rebinding needs connection-level pinning (resolve once, then connect to the validated IP while preserving SNI and the Host header), and I would rather land that as a focused, CI-verified change than rush it in here untested. Glad to fold it into this PR instead if you'd prefer it in scope.

**3. Missing `Location` header (medium).** In practice this could not `KeyError`: requests defines `is_redirect` as `"location" in self.headers and status in REDIRECT_STATI`, so the `headers['location']` read was only reachable once the header was present. I switched it to `.get('location')` with an explicit break regardless, so it is now safe by inspection.

One unrelated note for convenience: the red `unittest` job is 4 pre-existing import errors in other modules (`wandb.proto ... Imports`, `transformers.loss.loss_grounding_dino`) with `failures=0`. None of them touch `vision_utils`, and the new test is hermetic (no deps, ffmpeg, or network).

he-yufeng:
Another month on this one. The surface is unchanged on main: load_file still fetches whatever http(s) URL a template hands it, with no rejection of private or link-local hosts. The guard here stays narrow (block those, let everything else through). Happy to reshape it if you'd rather see it done differently.

tastelikefeet:
Thanks for your contribution! The issue you pointed out is indeed real. The main reason we're holding off on this PR is that we're currently in the middle of a large refactor, and the multimodal loading path is going to change. Once that lands, we'll reach out and invite you to re-evaluate this security issue.

he-yufeng:
Understood, that makes sense. I will keep this parked and re-verify the surface once the refactor lands. Thanks for the heads-up.

he-yufeng:
The deploy-boundary SSRF guard landed upstream in #10062/#10066 (merged 2026-09-08) per tastelikefeet's plan. Closing mine.

Issue #9745, opened 2026-07-14
Fix environment cleanup and rollout info handling issues in GRPO OpenEnv multi-turn scheduler


nantenT:
### Checklist / 检查清单

- [x] I have searched existing issues, and this is a new bug report. / 我已经搜索过现有的 issues，确认这是一个新的 bug report。

### Bug Description / Bug 描述

I found two issues while running the GRPO Sudoku example with OpenEnv:

### 1. `action_dict is None` path may lose `rollout_infos`

In `sudoku_scheduler.py` around line 280, when `action_dict` parsing fails, the code calls:

```python
await self._close_and_remove(uuid)
```

before returning the result.

After cleanup, related state associated with uuid may already be removed, causing subsequent access to uuid-related data in rollout_infos generation to fail.

A valid fix is to construct rollout_infos before calling _close_and_remove(uuid).

---

### 2. Environment leak when reaching `max_turns`

The example script:

```
examples/train/grpo/plugin/openenv/run_grpo_sudoku.sh
```

sets:

```bash
--max_turns 20
```

However, `sudoku_scheduler.py` currently only releases environments when the environment itself returns:

```python
done=True
```

via `on_turn_end()`.

When the rollout is terminated by the trainer because it reaches `max_turns`, the scheduler does not call `_close_and_remove()`. As a result, some OpenEnv sessions remain alive after the trajectory has already ended.

After enough rollouts, the number of active sessions keeps increasing until hitting the server limit (`max_sessions`), causing concurrency-related failures.

A cleanup hook for trajectory termination (or equivalent handling for `max_turns` termination) would prevent these leaked environments.

---

## Suggested fixes

* Move cleanup after `rollout_infos` construction in the `action_dict is None` branch.
* Add environment cleanup logic when a trajectory finishes due to `max_turns`, not only when `env.done=True`.
* Ideally provide a general trajectory-end callback/hook so environments can always be released regardless of how the rollout terminates.

### How to Reproduce / 如何复现

https://swift.readthedocs.io/zh-cn/latest/Instruction/GRPO/DeveloperGuide/gym_env.html#openenv

PR #9748, opened 2026-07-14
fix(examples/grpo): build rollout_infos before _close_and_remove in sudoku scheduler (#9745)

Anai-Guo:
## What

Fixes the first defect reported in #9745: in the OpenEnv Sudoku GRPO example, the parse-failure branch of `SudokuScheduler.on_turn_end` builds its `rollout_infos` dict **after** calling `_close_and_remove(uuid)`.

`OpenEnvScheduler._close_and_remove` (in `swift/rollout/multi_turn.py`) pops `self._total_rewards[uuid]` and `self._step_rewards[uuid]`. So when the model emits an unparseable action:

```python
self._total_rewards[uuid] = self._total_rewards.get(uuid, 0.0) - 1.0
self._step_rewards.setdefault(uuid, []).append(-1.0)
await self._close_and_remove(uuid)          # pops _total_rewards[uuid] / _step_rewards[uuid]
return {
    'done': True,
    'rollout_infos': {
        'total_reward': self._total_rewards[uuid],        # KeyError
        'step_rewards': list(self._step_rewards.get(uuid, [])),  # always []
        ...
```

`self._total_rewards[uuid]` raises `KeyError` (and `step_rewards` silently comes back empty). Since unparseable model output is common early in GRPO training, this crashes rollouts.

## Fix

Snapshot `rollout_infos` **before** cleanup — mirroring the already-correct terminal path at the end of `on_turn_end` (which builds `rollout_infos` first, then calls `_close_and_remove`). Minimal, behavior-preserving for the happy path.

## Note on the second reported issue

#9745 also reports an OpenEnv **session leak** when a trajectory stops because it hit `--max_turns` (or `finish_reason == 'length'`) rather than the env returning `done=True`. That one is a framework-level gap: `MultiTurnScheduler.run()` returns as soon as `should_stop` is set and never invokes a post-trajectory cleanup hook, so `on_turn_end` (the example's only entry point) has no way to release the env on a trainer-forced stop. Fixing it properly means adding a trajectory-end hook to the base scheduler and having `GYMScheduler`/`OpenEnvScheduler` close any surviving env — a change with broader blast radius across all schedulers, so I've left it out of this minimal fix and defer to maintainers on the preferred shape. Happy to follow up with that if you'd like.

### DIFF
--- a/examples/train/grpo/plugin/openenv/sudoku_scheduler.py
+++ b/examples/train/grpo/plugin/openenv/sudoku_scheduler.py
@@ -274,18 +274,20 @@ class SudokuScheduler(OpenEnvScheduler):
         action_text = response_choice.message.content
         action_dict = self.parse_action(action_text)
         if action_dict is None:
-            # Parse failed: end trajectory with penalty instead of polluting env
+            # Parse failed: end trajectory with penalty instead of polluting env.
+            # Snapshot rollout_infos *before* _close_and_remove, which pops
+            # self._total_rewards[uuid] / self._step_rewards[uuid] (see the
+            # base OpenEnvScheduler._close_and_remove). Building the dict after
+            # cleanup raised KeyError on total_reward and lost step_rewards.
             self._total_rewards[uuid] = self._total_rewards.get(uuid, 0.0) - 1.0
             self._step_rewards.setdefault(uuid, []).append(-1.0)
-            await self._close_and_remove(uuid)
-            return {
-                'done': True,
-                'rollout_infos': {
-                    'total_reward': self._total_rewards[uuid],
-                    'step_rewards': list(self._step_rewards.get(uuid, [])),
-                    'gym_done': True,
-                }
+            rollout_infos = {
+                'total_reward': self._total_rewards[uuid],
+                'step_rewards': list(self._step_rewards.get(uuid, [])),
+                'gym_done': True,
             }
+            await self._close_and_remove(uuid)
+            return {'done': True, 'rollout_infos': rollout_infos}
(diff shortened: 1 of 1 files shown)


nantenT:
Thanks for the fix! The current change correctly addresses the `rollout_infos` loss issue by building the information before calling `_close_and_remove`, which avoids the state cleanup ordering problem.

I also want to share a possible workaround for the second issue mentioned in #9745 (OpenEnv session leak when the trajectory ends due to `max_turns` / `length` instead of `env.done=True`).

A minimal scheduler-level workaround could be adding the external termination conditions into `on_turn_end`. For example, in the Sudoku scheduler:

```python
max_turns = 20

if done or current_turn >= max_turns - 1:
    await self._close_and_remove(uuid)
    done = True

return {
    "done": done,
    "rollout_infos": rollout_infos,
}
```

The same idea could be extended to `finish_reason == "length"` by tracking trajectory length/token usage inside the scheduler and treating it as another terminal condition.

This would provide a small, example-level workaround for users running multi-turn GRPO with OpenEnv, without requiring framework-level changes.

That said, I agree that the cleaner long-term solution would be introducing a scheduler lifecycle hook such as `on_trajectory_end(uuid, reason)`, because trajectory termination can happen from multiple sources:

* environment returns `done=True`
* `max_turns` reached
* generation stops due to `length`
* exception / trainer interruption

A unified lifecycle callback would make cleanup behavior consistent across different schedulers.

Thanks again for the fix!

hjh0119 (approved):

hjh0119:
thanks

PR #9750, opened 2026-07-15
fix(template): guard MiniCPM-V/mPLUG-Owl3 video sampling against 0-fps metadata

he-yufeng:
## What

`load_video_minicpmv_mplug_owl3` picks frames with

```python
sample_fps = round(vr.get_avg_fps() / 1)  # FPS
frame_idx = [i for i in range(0, len(vr), sample_fps)]
```

decord's `get_avg_fps()` returns `0.0` when the container has no / broken fps metadata (and a very low sub-1 fps rounds to 0 as well), so `sample_fps` becomes `0` and `range(0, len(vr), 0)` raises `ValueError: range() arg 3 must not be zero` before a single frame is read. The `MiniCPM-V` and `mPLUG-Owl3` templates call this for any user-supplied video (`swift/template/templates/minicpm.py`, `swift/template/templates/mplug.py`), and there is no upstream fps check, so an ordinary video with missing fps metadata crashes encoding. The same file already treats `fps == 0` as a real case elsewhere (`_video_get_metadata_local` uses `fps if fps > 0 else 0`).

## Fix

Clamp `sample_fps` to at least 1 so a 0-fps video falls back to sampling every frame, which the existing `max_num_frames` `uniform_sample` then downsamples. Normal videos (`fps >= 1`) are unaffected.

## Verification

decord/torch are not installable in my environment, so I reproduced the exact sampling logic standalone:

| avg fps | before | after |
| --- | --- | --- |
| 25.0 | 4 frames | 4 frames |
| 1.0 | 4 frames | 4 frames |
| 0.4 | `ValueError: range() arg 3 must not be zero` | 4 frames |
| 0.0 | `ValueError: range() arg 3 must not be zero` | 4 frames |

The added `test_load_video_minicpmv_handles_zero_fps` mocks a decord `VideoReader` whose `get_avg_fps()` returns `0.0` and asserts `load_video_minicpmv_mplug_owl3` returns frames instead of raising (fails before this change, passes after). It needs no real video or GPU.

### DIFF
--- a/swift/template/vision_utils.py
+++ b/swift/template/vision_utils.py
@@ -286,7 +286,7 @@ def load_video_minicpmv_mplug_owl3(video: Union[str, bytes], max_num_frames):
 
     video_io = load_file(video)
     vr = VideoReader(video_io, ctx=cpu(0))
-    sample_fps = round(vr.get_avg_fps() / 1)  # FPS
+    sample_fps = max(1, round(vr.get_avg_fps() / 1))  # FPS; clamp so a 0-fps video does not make range() step 0
     frame_idx = [i for i in range(0, len(vr), sample_fps)]
 
     if len(frame_idx) > max_num_frames:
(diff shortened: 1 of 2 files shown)


Jintao-Huang (approved):

Jintao-Huang:
thanks!

PR #9793, opened 2026-07-24
[Bugfix (rlhf)] skip model reload during colocate vLLM init on FSDP2

ys2025-AI:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

In the FSDP2 backend, GRPOTrainer.__init__ -> prepare_rollout() runs before accelerator.prepare() fully-shards the model, so self.model is still the full ~70GB. Exiting offload_context then load_model'd the full model back onto a single card and OOM'd at vLLM init.

Add a `reload` flag to RolloutTrainerMixin/GRPOTrainer/GKDTrainer offload_context (default True) and pass reload=False from _prepare_vllm at init, so the full model stays on CPU and is sharded+loaded later by train()'s accelerator.prepare(). Runtime rollouts keep reload=True (the model is now sharded, so reloading is cheap).

Also add qwen3_5 FSDP2 LoRA SFT/GRPO example scripts and an accelerate fsdp2.json config for Ascend.

### DIFF
--- a/examples/ascend/train/qwen3_5/fsdp2.json
+++ b/examples/ascend/train/qwen3_5/fsdp2.json
@@ -1,0 +1,22 @@
+{
+    "compute_environment": "LOCAL_MACHINE",
+    "debug": false,
+    "distributed_type": "FSDP",
+    "downcast_bf16": "no",
+    "mixed_precision": "bf16",
+    "num_machines": 1,
+    "num_processes": 8,
+    "machine_rank": 0,
+    "rdzv_backend": "static",
+    "same_network": true,
+    "use_cpu": false,
+    "fsdp_config": {
+      "fsdp_version": 2,
+      "fsdp_auto_wrap_policy": "TRANSFORMER_BASED_WRAP",
+      "fsdp_transformer_layer_cls_to_wrap": "Qwen3_5MoeDecoderLayer",
+      "fsdp_reshard_after_forward": true,
+      "fsdp_state_dict_type": "FULL_STATE_DICT",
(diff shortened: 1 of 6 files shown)


hjh0119:
Sorry for the late review, and thanks for digging into this!

Confirmed the issue is real — at prepare_rollout() time the model isn't sharded yet (accelerator.prepare happens later in _prepare_for_training()), so exiting offload_context() loads the full model onto one card.

Instead of adding a reload flag, just skip the offload context entirely in _prepare_vllm() when FSDP2 is enabled — the model is still on CPU at that point, so both the offload and the reload are pointless there. That avoids introducing an asymmetric context manager and keeps the logic in one place.

ys2025-AI:
> Sorry for the late review, and thanks for digging into this!
> 
> Confirmed the issue is real — at prepare_rollout() time the model isn't sharded yet (accelerator.prepare happens later in _prepare_for_training()), so exiting offload_context() loads the full model onto one card.
> 
> Instead of adding a reload flag, just skip the offload context entirely in _prepare_vllm() when FSDP2 is enabled — the model is still on CPU at that point, so both the offload and the reload are pointless there. That avoids introducing an asymmetric context manager and keeps the logic in one place.

Thanks for the review. I dropped the reload flag and instead skip the offload_context entirely in _prepare_vllm() when FSDP2 is enabled (rollout_mixin.py).

hjh0119 (approved):

hjh0119:
many thanks

Issue #9084, opened 2026-04-13
用swift微调Qwen3模型的function calling功能，如何构建数据集和模型参数以及如何避免过拟合和欠拟合


HGY19854338008:
### Checklist / 检查清单

- [x] I have searched existing issues, and this is a new question or discussion topic. / 我已经搜索过现有的 issues，确认这是一个新的问题与讨论。

### Question Description / 问题描述

这是数据集格式{"tools": [{"name": "query_ledger", "description": "查询电力台账数据。根据站点名称、母线名称、线路名称查询相关的电力设施台账信息。", "parameters": {"type": "object", "properties": {"subName": {"type": "string", "description": "站点名称"}}, "required": ["subName"]}}}], "messages": [{"role": "system", "content": "你是电网助手。先判断是否需要查询台账，再决定是否调用工具，最后给出结论。"}, {"role": "user", "content": "{\"alert_content\": \"2025-08-14 20:19:23 某某站 退出(前置机A报警)\", \"alert_level\": \"事故\", \"signal_id\": \"\", \"signal_package\": [\"2025-08-14 20:19:23 某某站 退出(前置机A报警)\"], \"station_inside_outer\": \"1\", \"throbNum\": \"0\"}"}, {"role": "assistant", "content": "<think>\n某某\"沙田站 退出(前置机A报警)\"描述关键字\"前置机A报警\"判断告警类型为四遥中断,四遥中断类型的信号需要调用 query_ledger 查询站点台账进行辅助判断信号的正确性。\n从告警内容中提取站点名称为\"某某站\",调用 query_ledger 工具查询某某站的台账信息。\n</think>", "tool_calls": [{"name": "query_ledger", "arguments": "{\"subName\": \"沙田站\", \"busName\": \"\", \"lineName\": \"\"}"}]}, {"role": "tool", "content": "[]"}, {"role": "assistant", "content": "<think>经查询某某站不在台账中，说明沙田站不属于调管厂站，不需要关注某某站的任何信号，而沙田站 退出(前置机A报警)告警信号属于某某站不在调管范围，所以这条信号为正常信号。\n</think>\n智能体查询到台账数量为0，说明台账中不存在相关某某站信息，说明某某站不属于调管厂站，不需要关注这个某某站告警信号，结果默认为正常信号。"}]}
这是模型参数/Users/e8900ai/anaconda3/envs/sft/bin/swift sft \
[...]

PR #9809, opened 2026-07-28
Support OpenAI and Anthropic message processors

lumincui:
## What changed

- normalize OpenAI `assistant.tool_calls` into SWIFT `tool_call` messages
- normalize Anthropic `tool_use` and `tool_result` content blocks into SWIFT canonical roles
- add explicit `OpenAIMessagesPreprocessor` and `AnthropicMessagesPreprocessor` classes
- auto-detect provider tool-call formats in `MessagesPreprocessor`
- document both accepted formats in English and Chinese

## Why

`MessagesPreprocessor._check_messages` keeps only `role`, `content`, `loss`, and `loss_scale`. OpenAI-style tool calls stored in `assistant.tool_calls` were therefore silently removed during dataset preprocessing even though templates can encode tool calls. This can produce training examples that retain assistant narration but lose the actual tool-call targets.

Anthropic tool-use traces similarly need their content blocks expanded into SWIFT's canonical `assistant`, `tool_call`, and `tool_response` roles before template encoding.

## Impact

OpenAI and Anthropic agent traces can now be passed through the normal dataset preprocessing path without an external conversion script. Parallel OpenAI tool calls, string or object arguments, Anthropic text/tool interleaving, nested text tool results, and loss metadata are covered.

The regression test also decodes supervised labels and asserts that the tool name is present, rather than only checking that template encoding succeeds.

## Validation

- `pre-commit run --all-files`
- `python -m unittest tests.general.test_data_preprocess.TestProviderMessagesPreprocess tests.general.test_data_preprocess.TestRejectedMessagesPreprocess tests.general.test_data_preprocess.TestDataPreprocess.test_tool_message`

Related to #8327 and #9084.

### DIFF
--- a/docs/source_en/Customization/Custom-dataset.md
+++ b/docs/source_en/Customization/Custom-dataset.md
@@ -11,7 +11,7 @@ The following is an introduction to the dataset formats that `AutoPreprocessor` 
 
 The standard dataset format for ms-swift accepts keys such as: 'messages', 'rejected_response', 'label', 'images', 'videos', 'audios', 'tools', and 'objects'. Among these, 'messages' is a required key. 'rejected_response' is used for DPO and other RLHF training, 'label' is used for KTO training and classification model training. The keys 'images', 'videos', and 'audios' are used to store paths or URLs for multimodal data, 'tools' is used for Agent tasks, and 'objects' is used for grounding tasks.
 
-There are three core preprocessors in ms-swift: `MessagesPreprocessor`, `AlpacaPreprocessor`, and `ResponsePreprocessor`. `MessagesPreprocessor` is used to convert datasets in the messages and sharegpt format into the standard format. `AlpacaPreprocessor` converts datasets in the alpaca format, while `ResponsePreprocessor` converts datasets in the query/response format. `AutoPreprocessor` automatically selects the appropriate preprocessor for the task.
+There are three core preprocessors in ms-swift: `MessagesPreprocessor`, `AlpacaPreprocessor`, and `ResponsePreprocessor`. `MessagesPreprocessor` is used to convert datasets in the messages and sharegpt format into the standard format. It also automatically normalizes OpenAI `tool_calls` and Anthropic `tool_use`/`tool_result` content blocks. The provider-specific `OpenAIMessagesPreprocessor` and `AnthropicMessagesPreprocessor` classes can be used when explicit format selection is preferred. `AlpacaPreprocessor` converts datasets in the alpaca format, while `ResponsePreprocessor` converts datasets in the query/response format. `AutoPreprocessor` automatically selects the appropriate preprocessor for the task.
 
 The following four formats will all be converted into the `messages` field of the ms-swift standard format under the processing of `AutoPreprocessor`, meaning they can all be directly used with `--dataset <dataset-path>`:
 
@@ -24,6 +24,16 @@ Messages format (standard format):
 ShareGPT format:
 ```jsonl
 {"system": "<system>", "conversation": [{"human": "<query1>", "assistant": "<response1>"}, {"human": "<query2>", "assistant": "<response2>"}]}
+```
+
+OpenAI tool-call format:
+```jsonl
+{"messages": [{"role": "user", "content": "How is the weather?"}, {"role": "assistant", "content": null, "tool_calls": [{"type": "function", "function": {"name": "get_weather", "arguments": "{\"city\":\"Beijing\"}"}}]}, {"role": "tool", "content": "Sunny"}]}
+```
+
+Anthropic tool-use format:
+```jsonl
+{"messages": [{"role": "assistant", "content": [{"type": "text", "text": "I will check."}, {"type": "tool_use", "id": "toolu_01", "name": "get_weather", "input": {"city": "Beijing"}}]}, {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "toolu_01", "content": "Sunny"}]}]}
 ```
 
 Query-Response format:
--- a/docs/source_en/Customization/Custom-dataset.md
+++ b/docs/source_en/Customization/Custom-dataset.md
@@ -31,6 +31,12 @@ OpenAI tool-call format:
 {"messages": [{"role": "user", "content": "How is the weather?"}, {"role": "assistant", "content": null, "tool_calls": [{"type": "function", "function": {"name": "get_weather", "arguments": "{\"city\":\"Beijing\"}"}}]}, {"role": "tool", "content": "Sunny"}]}
 ```
 
+OpenAI Chat Completions multimodal tool-call format:
+```jsonl
+{"messages": [{"role": "user", "content": [{"type": "text", "text": "Compare these images."}, {"type": "image_url", "image_url": {"url": "data:image/png;base64,{base64_encoded}"}}, {"type": "image_url", "image_url": "https://example.com/input.png"}]}, {"role": "assistant", "content": [{"type": "text", "text": "I will inspect them."}], "tool_calls": [{"type": "function", "function": {"name": "inspect_images", "arguments": "{\"detail\":\"high\"}"}}]}]}
+```
+- OpenAI `image_url` blocks are converted to `<image>` placeholders and the top-level `images` field in their original order. This example uses the Chat Completions format; Responses API `input_text` and `input_image` blocks use a different schema.
+
 Anthropic tool-use format:
 ```jsonl
 {"messages": [{"role": "assistant", "content": [{"type": "text", "text": "I will check."}, {"type": "tool_use", "id": "toolu_01", "name": "get_weather", "input": {"city": "Beijing"}}]}, {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "toolu_01", "content": "Sunny"}]}]}
--- a/docs/source_en/Customization/Custom-dataset.md
+++ b/docs/source_en/Customization/Custom-dataset.md
@@ -11,9 +11,9 @@ The following is an introduction to the dataset formats that `AutoPreprocessor` 
 
 The standard dataset format for ms-swift accepts keys such as: 'messages', 'rejected_response', 'label', 'images', 'videos', 'audios', 'tools', and 'objects'. Among these, 'messages' is a required key. 'rejected_response' is used for DPO and other RLHF training, 'label' is used for KTO training and classification model training. The keys 'images', 'videos', and 'audios' are used to store paths or URLs for multimodal data, 'tools' is used for Agent tasks, and 'objects' is used for grounding tasks.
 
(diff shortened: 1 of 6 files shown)


vincenthcui:
Thanks @ezeli and @HGY19854338008 for raising #8327 and #9084 and providing concrete OpenAI tool-call examples — they were very helpful when shaping this change.

This PR adds normalization for OpenAI `assistant.tool_calls` and Anthropic tool-use content blocks while continuing to use the existing `MessagesPreprocessor` pipeline. @ezeli and @HGY19854338008, if you have time, it would be great if you could check whether the OpenAI behavior matches what you had in mind.

@Jintao-Huang @hjh0119 @tastelikefeet, when convenient, we would also appreciate your guidance on whether `MessagesPreprocessor` is the right integration point for these provider formats. Happy to adjust the design based on your feedback. Thank you!

swift/dataset/preprocessor/core.py
```diff
@@ -507,6 +510,112 @@ def to_std_messages(self, messages: List[Dict[str, str]], system: Optional[str])
...
+        new_messages = []
+        for message in messages:
+            if message.get('role') != 'assistant' or not message.get('tool_calls'):
+                new_messages.append(message)
+                continue
+
+            content = message.get('content')
+            if content:
+                new_messages.append({
+                    key: value
```
ezeli, line 534: message["content"] = f'{message["content"]}\n\n'
'\n\n' are needed to maintain consistency.

swift/dataset/preprocessor/core.py
```diff
@@ -507,6 +510,112 @@ def to_std_messages(self, messages: List[Dict[str, str]], system: Optional[str])
...
+                    for key, value in message.items() if key in {'role', 'content', 'loss', 'loss_scale'}
+                })
+            for tool_call in message['tool_calls']:
```
ezeli, line 537: My implementation:
```python
for tool_call in message['tool_calls']:
    func = tool_call.get('function', tool_call)
    arguments = func.get('arguments', {})
    if isinstance(arguments, str):
        arguments = json.loads(arguments)
    for arg_k, arg_v in arguments.items():
        if isinstance(arg_v, (dict, list)):
            arguments[arg_k] = json.dumps(arg_v, ensure_ascii=False)
    func['arguments'] = arguments
    func_content = json.dumps(func, ensure_ascii=False)
    new_messages.append({'role': 'tool_call', 'content': func_content})
```

tastelikefeet:
Thanks! I will check this code asap

swift/dataset/preprocessor/core.py
vincenthcui: Thanks for pointing this out — I agree that the encoded output should preserve the same `\n\n` separator as the inference chat template.

On current `main`, this separator is added in the model-formatting layer by `Qwen3_5AgentTemplate._add_tool_call_prefix()` when a non-empty assistant message precedes a tool call. Adding it here in the provider preprocessor would make the OpenAI normalization Qwen-specific and could also produce duplicate separators for Qwen3.5/3.6.

So I think it is safer for this processor to preserve the assistant content as-is and let each agent template add its required separator. A focused encode-parity test can make this behavior explicit if you think that would be useful. Thanks!

swift/dataset/preprocessor/core.py
vincenthcui: Thanks for sharing your implementation — it helped clarify the expected behavior.

The PR intentionally keeps the normalized tool-call payload structured as `{"name": ..., "arguments": ...}`. The existing `BaseAgentTemplate._parse_tool_call()` accepts either a dict or a JSON string, and model-specific formatters handle serialization afterward. For example, the Qwen3 Coder formatter already applies `json.dumps()` to nested dict/list argument values.

Serializing nested values in the provider preprocessor would lose their original types and may lead to double encoding in templates such as Hermes. Keeping the arguments structured also makes the normalized representation consistent with ms-swift's canonical `tool_call` format.

A nested dict/list argument case can also be added to make this distinction explicit. If you have a template where the structured form produces different encoded output, a small example would be very helpful. Thank you!

vincenthcui: I added a regression test in b43130f for nested list/dict arguments.

The test verifies both stages:

1. OpenAI JSON arguments are normalized to the canonical ms-swift message:
   `{"role": "tool_call", "content": {"name": ..., "arguments": {...}}}`, with nested values still typed as lists/dicts.
2. After the existing template encodes that canonical message, the tool name and nested argument values are present in supervised labels.

This keeps provider normalization lossless while leaving model-specific serialization to the agent template. The focused pre-commit checks and related unit tests pass.

vincenthcui: I tested the suggested implementation locally with this nested-argument case:

```json
{
  "name": "get_weather",
  "arguments": {
    "cities": ["Beijing", "Shanghai"],
    "options": {"units": ["celsius", "fahrenheit"]}
  }
}
```

Both representations can be encoded successfully by the Hermes template, and the tool name and argument values reach supervised labels. The difference appears before template encoding:

```python
# Structured canonical message used by this PR
{
    'role': 'tool_call',
    'content': {
        'name': 'get_weather',
        'arguments': {
            'cities': ['Beijing', 'Shanghai'],
            'options': {'units': ['celsius', 'fahrenheit']},
        },
    },
}
```

With the suggested early serialization, `content` becomes a JSON string and the nested values become strings as well:

```python
{
    'cities': '["Beijing", "Shanghai"]',
    'options': '{"units": ["celsius", "fahrenheit"]}',
}
```

The exact suggested snippet therefore does not pass the PR's canonical-structure tests, although template encoding itself succeeds after adapting the assertions to the serialized representation. It would also need to copy `loss` and `loss_scale` from the source assistant message.

This is why the PR keeps the canonical message structured and lossless, then lets the existing agent template perform model-specific serialization. The new regression test in b43130f covers this case. Thanks again for sharing the alternative implementation — it was useful for checking the boundary explicitly.

swift/dataset/preprocessor/core.py
ezeli: `Qwen3_5AgentTemplate._add_tool_call_prefix()`was probably added later. I just found out, thank you.

swift/dataset/preprocessor/core.py
ezeli: https://github.com/modelscope/ms-swift/blob/1d64b1bc4a18de1a28545a798a69dc999342620d/swift/agent_template/qwen3_coder.py#L146

Previously, the parameter `ensure_ascii=False` wasn't present, and I mainly used it to solve this problem. However, now that it's been added, it's no longer needed.

swift/dataset/preprocessor/core.py
vincenthcui: Thanks for confirming, and thank you again for catching the consistency point. It is good that the current template now handles the separator in the model-specific layer.

swift/dataset/preprocessor/core.py
vincenthcui: Thanks for the context — that explains the earlier serialization approach. Since the current formatter now applies `ensure_ascii=False`, we can keep the provider normalization structured and let the template handle the final serialization. I appreciate you checking this with us!

swift/dataset/preprocessor/core.py
```diff
@@ -507,6 +510,112 @@ def to_std_messages(self, messages: List[Dict[str, str]], system: Optional[str])
...
+                            'arguments': block.get('input', {}),
+                        },
+                        **message_metadata,
+                    })
+                elif block_type == 'tool_result':
+                    flush_text()
+                    new_messages.append({
+                        'role': 'tool_response',
+                        'content': cls._anthropic_block_content(block.get('content', '')),
+                        **message_metadata,
```
tastelikefeet, line 632: What about the multi-modal blocks?

docs/source_en/Customization/Custom-dataset.md
```diff
@@ -26,6 +26,16 @@ ShareGPT format:
 {"system": "<system>", "conversation": [{"human": "<query1>", "assistant": "<response1>"}, {"human": "<query2>", "assistant": "<response2>"}]}
 ```
 
+OpenAI tool-call format:
+```jsonl
+{"messages": [{"role": "user", "content": "How is the weather?"}, {"role": "assistant", "content": null, "tool_calls": [{"type": "function", "function": {"name": "get_weather", "arguments": "{\"city\":\"Beijing\"}"}}]}, {"role": "tool", "content": "Sunny"}]}
```
tastelikefeet, line 31: An example of multi-modal messages is needed

swift/dataset/preprocessor/core.py
vincenthcui: Thank you for catching this. The previous implementation only handled text and tool blocks, so Anthropic image blocks could be silently dropped.

I addressed this in 526c4123:

- base64 and URL image sources are converted to ms-swift's canonical `<image>` placeholders plus the top-level `images` field, preserving their original order;
- the same conversion works for both regular message content and multimodal `tool_result` content;
- unsupported image source/content block types now raise a clear error instead of being silently discarded;
- auto-detection also recognizes Anthropic image blocks without requiring a tool block.

The regression test covers a mixed `text -> base64 image -> tool_use -> URL image tool_result -> text` conversation, verifies the canonical messages and ordered `images` values, checks `StdTemplateInputs` compatibility, and loads the base64 image through ms-swift's existing `load_image` path. Thank you for the helpful review!

docs/source_en/Customization/Custom-dataset.md
vincenthcui: Thanks — I added matching Anthropic multimodal tool-use examples to both the English and Chinese custom-dataset documentation in 526c4123.

The example includes a base64 image in a normal user message and a URL image inside `tool_result`, and explains that they are converted to `<image>` placeholders and the top-level `images` field in their original order.

swift/dataset/preprocessor/core.py
vincenthcui: As a follow-up, I also added symmetric regression coverage for OpenAI Chat Completions multimodal messages in `54b24c99`. No additional OpenAI processor conversion was needed because the existing `StdTemplateInputs.remove_messages_media` path already extracts `image_url` blocks into `<image>` placeholders and the top-level `images` field. The new test verifies base64 and URL images in order, real base64 decoding, and that assistant text can coexist with normalized `tool_calls`. The focused 9-test suite and all pre-commit hooks pass. Responses API `input_text`/`input_image` blocks use a separate schema and are documented as outside this Chat Completions example. Thank you again for prompting the multimodal coverage!

docs/source_en/Customization/Custom-dataset.md
vincenthcui: I have now added matching OpenAI Chat Completions multimodal examples to both the English and Chinese documentation in `54b24c99`. The example covers base64 and URL `image_url` blocks together with an assistant `tool_calls` entry, and documents their conversion to ordered `<image>` placeholders plus the top-level `images` field. It also clarifies that Responses API `input_text`/`input_image` blocks use a different schema. Thanks again for the helpful suggestion!

vincenthcui:
Hi @tastelikefeet, thank you again for the helpful review. The requested multimodal handling, documentation examples, and regression coverage have now been added, and the focused tests and pre-commit checks pass.

When convenient, would you mind taking another look? We would really appreciate having this merged into `main` when it is ready, as our system is currently carrying these changes as a downstream patch version. We are happy to make any further adjustments you think are needed. Thank you!

tastelikefeet (approved):

PR #9832, opened 2026-08-01
fix: preserve explicit PEFT LoRA dtype

### LINKED ISSUE #9694: `lora_dtype` has no effect with peft backend — LoRA checkpoint always saved as float32
### Checklist / 检查清单

### Bug Description / Bug 描述

# `lora_dtype` has no effect with peft backend — LoRA checkpoint always saved as float32

## Summary

Setting `--lora_dtype bfloat16` (or `float16`) has no effect when using the **peft backend** (the default). LoRA weights are always cast to `float32` regardless of the `lora_dtype` value. This affects both the in-memory LoRA parameters and the saved checkpoint.

The root cause is that peft's `_cast_adapter_dtype(autocast_adapter_dtype=True)` (the default) runs **after** swift's `_convert_dtype`, silently upcasting bf16/fp16 back to float32.

## Environment

| Component | Version |
|---|---|
| ms-swift | 4.0.3 (also confirmed on 4.3.2) |
| peft | 0.18.1 (also confirmed on 0.19.1) |
| Docker image | `modelscope-registry.us-west-1.cr.aliyuncs.com/modelscope-repo/modelscope:ubuntu22.04-cuda12.8.1-py311-torch2.10.0-vllm0.17.1-modelscope1.34.0-swift4.0.3` |

## Reproduction

Minimal Python repro (no training needed):

```python
import torch
from swift.tuners.peft import LoraConfig as SwiftLoraConfig
from peft import get_peft_model

class MiniModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = torch.nn.Linear(32, 64, bias=False)
    def forward(self, x):
        return self.linear(x)

# Default behavior (autocast_adapter_dtype=True, what swift uses internally)
model1 = MiniModel()
lc1 = SwiftLoraConfig(r=8, lora_alpha=16, target_modules=["linear"], lora_dtype="bfloat16")
pm1 = get_peft_model(model1, lc1, autocast_adapter_dtype=True)
for n, p in pm1.named_parameters():
    if "lora" in n:
        print(f"  {n}: {p.dtype}")
# → lora_A.default.weight: torch.float32  ← should be bfloat16
# → lora_B.default.weight: torch.float32  ← should be bfloat16

# With autocast_adapter_dtype=False (not exposed by swift CLI)
model2 = MiniModel()
lc2 = SwiftLoraConfig(r=8, lora_alpha=16, target_modules=["linear"], lora_dtype="bfloat16")
pm2 = get_peft_model(model2, lc2, autocast_adapter_dtype=False)
for n, p in pm2.named_parameters():
    if "lora" in n:
        print(f"  {n}: {p.dtype}")
# → lora_A.default.weight: torch.bfloat16  ✅
# → lora_B.default.weight: torch.bfloat16  ✅
```

Command-line that triggers the bug:

```bash
swift sft \
    --model /path/to/model \
    --tuner_type lora \
    --lora_dtype bfloat16 \
    --torch_dtype bfloat16 \
    ...
```

### Saved checkpoint verification

```python
from safetensors import safe_open

with safe_open("checkpoint-xxx/adapter_model.safetensors", framework="pt") as f:
    for k in list(f.keys())[:4]:
        t = f.get_tensor(k)
        print(k, t.dtype)
# base_model.model...lora_A.weight: dtype=torch.float32
# base_model.model...lora_B.weight: dtype=torch.float32
```

```json
// additional_config.json (swift records the config, but it didn't take effect)
{"lora_dtype": "bfloat16", "lorap_lr_ratio": null, "lorap_emb_lr": 1e-06}
```

### Call order in `PeftModel.__init__`

`peft/peft_model.py`

```python
# Line 127-133
self.base_model = cls(model, {adapter_name: peft_config}, adapter_name)
#             ^^^^ = LoraModel.__init__ = swift's __new_init__ (monkey-patched)

if hasattr(self.base_model, "_cast_adapter_dtype"):
    self.base_model._cast_adapter_dtype(
        adapter_name=adapter_name,
        autocast_adapter_dtype=autocast_adapter_dtype,  # ← default True
    )
```

### swift's `__new_init__` runs `_convert_dtype` too early

`swift/tuners/peft.py`

```python
# Line 340-356
def __new_init__(self, model, config, *args, **kwargs):
    self.__init_origin__(model, config, *args, **kwargs)  # Step 1: peft inject_adapter → float32
    ...
    for active_adapter in active_adapters:
        active_config = config[active_adapter] if isinstance(config, dict) else config
        if hasattr(active_config, 'lora_dtype'):
            for name, module in model.named_modules():
                if isinstance(module, LoraLayer):
                    _convert_dtype(module, active_adapter, active_config.lora_dtype)
                    # ^^^ Step 2: .to(bfloat16) → dtype becomes bf16 ✅
```

### peft's `_cast_adapter_dtype` overrides it immediately after

`peft/tuners/tuners_utils.py`

```python
# Line 1976-2016
def cast_adapter_dtype(model, adapter_name, autocast_adapter_dtype=True):
    """Currently, this only upcasts float16 and bfloat16 to float32."""
    if not autocast_adapter_dtype:
        return
    dtypes_to_convert_to_fp32 = {torch.float16, torch.bfloat16}
    ...
    for param in submodule[adapter_name].parameters():
        if param.dtype in dtypes_to_convert_to_fp32:
            param.data = param.data.to(torch.float32)
            # ^^^ Step 3: bf16 → float32 ❌ overrides swift's _convert_dtype
```

### Timeline

```
PeftModel.__init__
 ├─ 1. LoraModel.__init__ (= swift __new_init__)
 │   ├─ __init_origin__ → inject_adapter        → float32  (peft default)
 │   └─ _convert_dtype(lora_dtype="bfloat16")   → bfloat16 ✅  (swift works)
 │
 └─ 2. _cast_adapter_dtype(autocast=True)       → float32  ❌  (peft overrides)
      └─ cast_adapter_dtype: "upcasts float16 and bfloat16 to float32"
```

`lora_dtype` is a swift-only extension (peft has no such field — it's popped in `to_peft_config()`), so peft has no way to know it should be respected.

## Evidence that `_convert_dtype` works but gets overridden

With a debug-instrumented `_convert_dtype`:

```
=== Calling get_peft_model ===
  [_convert_dtype CALLED] adapter=default, lora_dtype=bfloat16, target=Linear
    BEFORE: lora_A dtype = torch.float32
    AFTER:  lora_A dtype = torch.bfloat16      ← swift's _convert_dtype succeeded
=== _convert_dtype was called 1 times ===

=== Final LoRA dtypes ===
  lora_A.default.weight: torch.float32          ← but peft's _cast_adapter_dtype ran after
  lora_B.default.weight: torch.float32
```

## Impact

- `--lora_dtype bfloat16` / `float16` is silently ignored with the peft backend (default)
- Checkpoint size is 2x larger than expected (fp32 vs bf16)
- Users cannot control LoRA dtype as documented
- `additional_config.json` records `lora_dtype: bfloat16` but the actual tensors are float32 — misleading

### Option A: Pass `autocast_adapter_dtype=False` when `lora_dtype` is set

`swift/pipelines/train/tuner.py`

```python
# When lora_dtype is specified, disable peft's auto-upcast
pm = get_peft_model(model, lora_config, autocast_adapter_dtype=(args.lora_dtype is None))
```

### Option B: Move `_convert_dtype` to run after `_cast_adapter_dtype`

Patch `PeftModel.__init__` or hook into a point after `_cast_adapter_dtype` completes.

### Option C: Cast dtype on save

In the save path, convert LoRA parameters to `lora_dtype` before writing to disk.

### How to Reproduce / 如何复现

ENABLE_AUDIO_OUTPUT=0 \
IMAGE_MAX_TOKEN_NUM=1024 \
VIDEO_MAX_TOKEN_NUM=128 \
FPS_MAX_FRAMES=12 \
NPROC_PER_NODE=2 \
MASTER_PORT=${MASTER_PORT} \
CUDA_VISIBLE_DEVICES=2,3 \
swift sft \
    --model "Qwen3-Omni-30B-A3B-Instruct" \
    --dataset 'xxx \
    --load_from_cache_file true \
    --tuner_type lora \
    --torch_dtype bfloat16 \
    --num_train_epochs 2 \
    --per_device_train_batch_size 1 \
    --per_device_eval_batch_size 1 \
    --router_aux_loss_coef 1e-3 \
    --attn_impl flash_attn \
    --learning_rate 1e-4 \
    --lora_rank 8 \
    --lora_alpha 32 \
    --lora_dtype bfloat16 \
    --freeze_vit true \
    --freeze_aligner true \
    --gradient_accumulation_steps 1 \
    --eval_steps 1000 \
    --save_steps 10000 \
    --save_total_limit 2 \
    --logging_steps 1 \
    --max_length 8192 \
    --output_dir "${OUTPUT_DIR}" \
    --warmup_ratio 0.05 \
    --dataset_num_proc 1 \
    --lora_dropout 0 \
    --target_parameters gate_up_proj down_proj gate.weight \
    --dataloader_num_workers 4

_No response_

yaodong-shen:
## What does this PR do?

Fixes #9694. When the PEFT backend receives Swift's explicit `lora_dtype` (`float16`/`bfloat16`), PEFT's post-initialization `_cast_adapter_dtype` currently upcasts those adapter weights back to fp32. This makes the documented Swift option ineffective and produces unnecessarily large checkpoints.

This patch hooks that cast only when the Swift `LoraConfig` explicitly has `lora_dtype`, preserving PEFT's default behavior for all other adapters. The existing skipped `test_peft_lora_dtype` regression test is enabled and now checks both initial injection and reload preserve fp16.

## Validation

- `python3 -m py_compile swift/tuners/peft.py tests/tuners/test_peft.py`
- `git diff --check` on both changed files
- Full PEFT test requires the repository CI dependencies (`torch`, `peft`, ModelScope); they are not installed in this environment.

### DIFF
--- a/swift/tuners/peft.py
+++ b/swift/tuners/peft.py
@@ -144,6 +144,22 @@ def _convert_dtype(target: torch.nn.Module, adapter_name: str, lora_dtype: str):
             target.lora_embedding_B[adapter_name].to(torch_dtype)
 
 
+def _cast_adapter_dtype_hook(self, adapter_name: str, autocast_adapter_dtype: bool = True):
+    """Keep an explicitly configured Swift LoRA dtype from being upcast by PEFT.
+
+    PEFT calls ``_cast_adapter_dtype`` after the LoRA modules have been
+    created.  Its default behavior upcasts fp16/bf16 adapters to fp32, which
+    silently overrides Swift's ``lora_dtype`` setting.  Only disable that
+    automatic cast when the Swift extension is explicitly configured; the
+    default PEFT behavior remains unchanged for all other adapters.
+    """
+    peft_config = getattr(self, 'peft_config', {})
+    config = peft_config.get(adapter_name) if isinstance(peft_config, dict) else None
+    if getattr(config, 'lora_dtype', None) is not None:
+        autocast_adapter_dtype = False
+    return self._cast_adapter_dtype_origin(adapter_name, autocast_adapter_dtype)
+
+
 def create_optimizer_param_groups(self: PeftModel, **defaults):
     if not isinstance(self.peft_config[self.active_adapter],
                       LoraConfig) or self.peft_config[self.active_adapter].lorap_lr_ratio is None:
@@ -361,6 +377,9 @@ def hot_patch_peft_module():
 
     LoraModel.__init_origin__ = LoraModel.__init__
     LoraModel.__init__ = __new_init__
+    if not hasattr(LoraModel, '_cast_adapter_dtype_origin'):
+        LoraModel._cast_adapter_dtype_origin = LoraModel._cast_adapter_dtype
+        LoraModel._cast_adapter_dtype = _cast_adapter_dtype_hook
 
     # Support LoRA+
     PeftModel.create_optimizer_param_groups = create_optimizer_param_groups
--- a/tests/tuners/test_peft.py
+++ b/tests/tuners/test_peft.py
@@ -133,18 +133,19 @@ class TestPeft(unittest.TestCase):
             self.assertTrue(key in state_dict2)
             self.assertTrue(all(torch.isclose(state_dict[key], state_dict2[key]).flatten().detach().cpu()))
 
-    @unittest.skip
     def test_peft_lora_dtype(self):
         model = SbertForSequenceClassification(SbertConfig())
         model2 = copy.deepcopy(model)
         model3 = copy.deepcopy(model)
         lora_config = LoraConfig(target_modules=['query', 'key', 'value'], lora_dtype='float16')
         model = Swift.prepare_model(model, lora_config)
+        self.assertTrue(model.base_model.bert.encoder.layer[0].attention.self.key.lora_A.default.weight.dtype ==
+                        torch.float16)
         model.save_pretrained(self.tmp_dir, safe_serialization=False)
         self.assertTrue(os.path.exists(os.path.join(self.tmp_dir, 'additional_config.json')))
         model2 = Swift.from_pretrained(model2, self.tmp_dir)
         self.assertTrue(model2.base_model.model.bert.encoder.layer[0].attention.self.key.lora_A.default.weight.dtype ==
-                        torch.float32)
+                        torch.float16)
         self.assertTrue(model2.peft_config['default'].lora_dtype == 'float16')
         state_dict = model.state_dict()
         state_dict2 = model2.state_dict()


ErenAta16 (commented):
The premise checks out against PEFT rather than just against the symptom. `_cast_adapter_dtype` says so in its own docstring:

```
peft 0.19.1
signature: (self, adapter_name: str, autocast_adapter_dtype: bool = True) -> None

    A helper method to cast the adapter weights to the correct dtype.
    Currently, this only upcasts float16 and bfloat16 to float32.
```

So an explicitly configured fp16/bf16 `lora_dtype` really is overridden after the modules are built, and the signature the hook wraps matches exactly, including the default.

The scoping is also right, and this is the part worth stating because it is not obvious from the diff. `_cast_adapter_dtype` is defined on `BaseTuner`, not on `LoraModel`:

```
defined on: BaseTuner  (peft.tuners.tuners_utils)
LoraModel.__mro__: LoraModel -> BaseTuner -> Module -> ABC -> object
```

so assigning to `LoraModel._cast_adapter_dtype` shadows the inherited method on the LoRA branch only. Tuners that inherit `BaseTuner` directly keep PEFT's behaviour untouched, and LoRA variants that subclass `LoraModel` pick the hook up, which is what you want given `lora_dtype` is a LoRA-side setting. Gating on `getattr(config, 'lora_dtype', None) is not None` keeps it to adapters Swift actually configured, so a plain PEFT adapter in the same process is unaffected.

Two things worth a look.

**The idempotency guard protects against double-wrapping but not against a silent no-op.** `if not hasattr(LoraModel, '_cast_adapter_dtype_origin')` is the right shape for repeated `__new_init__` calls. The failure mode I would want covered is a future PEFT that renames or moves this private method: `LoraModel._cast_adapter_dtype_origin = LoraModel._cast_adapter_dtype` would raise `AttributeError` at that point, which is loud and fine, but only if that line is not inside a `try`. Worth confirming it is not, because a swallowed exception there means the patch quietly stops applying and `lora_dtype` starts being overridden again with no signal, which is exactly the bug this fixes.

**Explicit fp16 against an fp32 base.** Disabling the autocast for a configured `lora_dtype` is correct as an expression of user intent, but it means someone setting `lora_dtype='float16'` on a model loaded in fp32 now gets an adapter whose dtype differs from the base weights, where before PEFT quietly harmonised them. That is presumably the point, so I would not change it; a line in the docstring saying the cast is skipped entirely rather than adjusted would save the next reader from wondering whether a mismatch is a bug.

Pinning a private API is inherent to the approach and I do not think there is a better option here, but a comment naming the PEFT version this was verified against would give whoever hits a future break a starting point.

yaodong-shen:
Thanks for the detailed review and for checking the PEFT implementation details.

I confirmed both points:

- The `_cast_adapter_dtype_origin` lookup and hook assignment are outside any `try` block, so a future PEFT rename/removal fails loudly during patch setup rather than silently disabling the fix.
- `test_peft_lora_dtype` uses the default Sbert base (fp32) with an explicit `lora_dtype='float16'` and asserts the injected/reloaded LoRA weights remain fp16. That mismatch is intentional: the Swift setting is an explicit adapter-dtype request, while the hook only disables PEFT's automatic upcast for configured Swift adapters.

The hook docstring also documents that the automatic cast is skipped only for explicitly configured Swift LoRA adapters; plain PEFT adapters retain PEFT's default behavior. No additional code change is needed from these observations.

tastelikefeet (approved):

PR #9833, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
Clarify that the OpenAI Python client `base_url` pattern works with ms-swift and with other OpenAI-compatible multi-model gateways, using DaoXE (https://api.daoxe.com/v1) as one concrete example. Docs only.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


seven7763:
It's a generic note that says nothing about ms-swift in particular, and the tracked link makes it look like exactly what it was. It comes down. Sorry for adding it to your queue.

PR #9844, opened 2026-08-04
Fix GRPO offload compatibility with FSDP2 CPU offload

addsubmuldiv:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation Updates
- [ ] Support for More Models or Datasets

# PR information

## Background

When using Qwen3.5 for GRPO training on Ascend/NPUs, the combination of activation offload, FSDP2 parameter CPU offload, and `--vllm_enable_lora True` has compatibility issues. These can cause training to fail during rollout, log-probability computation, or the backward pass. The same issues were also reproduced on NVIDIA A100 GPUs under the equivalent GRPO configuration.

### 1. Fix incorrect activation CPU offload state during no-grad forwards

GRPO performs `torch.no_grad()` forwards while the model is in `model.train()` mode when computing old/reference log-probabilities. Previously, the activation offload wrapper still advanced the offload handler's forward state, but these forwards have no corresponding backward pass. This could corrupt subsequent activation-group state and trigger a `KeyError`.

This change:

- Bypasses the activation offload handler for forwards in non-training mode or under `torch.no_grad()`;
- Uses `try/finally` to ensure that the handler's post-forward cleanup always runs.

#### 2.1 Fix device selection for GRPO batches

With FSDP2 parameter CPU offload enabled, `model.device` may report CPU even though actual training computation runs on the current NPU/GPU.

Previously, GRPO used `model.device` when constructing batches for:

- Regular policy batches;
- OPSD teacher batches;
- Re-encoded batches for dynamic multimodal chunks.

This could place `input_ids`, masks, and other GRPO batch tensors on CPU while model forward outputs were on the NPU/GPU. Later, when computing old/reference log-probabilities, `gather` was called using CPU `input_ids` and NPU/GPU logits, causing device mismatch errors.

This change consistently uses `self.accelerator.device` when constructing these batches, ensuring model inputs and GRPO auxiliary tensors are placed on the actual computation device.

#### 2.2 Fix LoRA parameter synchronization for CPU DTensors

With FSDP2 parameter CPU offload enabled, LoRA parameter states may be returned as CPU DTensors. Previously, calling `full_tensor()` directly could cause HCCL/NPU collectives to run on CPU DTensors, resulting in communication backend errors because CPU is unsupported.

This change detects CPU DTensors before calling `full_tensor()`, moves them to the current computation device, and then performs the full-tensor gather.

#### 2.3 Fix global gradient norm computation for CPU-offloaded gradients

With FSDP2 parameter CPU offload enabled, parameter gradients may reside on CPU. Accelerate's default global gradient norm path directly computes norms and collectives on these CPU gradients/DTensors, causing CPU DTensor device errors.

For the FSDP2 case where gradients are on CPU, this change:

- Unscales gradients first;
- Uses `foreach_norm` to compute individual gradient norms;
- Moves norm tensors to the current computation device;
- Computes the global norm and applies gradient clipping.

Non-FSDP2 scenarios, or scenarios without CPU-offloaded gradients, continue to use the original `Accelerator.clip_grad_norm_` implementation.

## Validation

- Verified staged changes with `git diff --cached --check`;
- Ran end-to-end smoke tests on Ascend 910B3 NPUs and NVIDIA A100 GPUs using Qwen3.5-4B, 2 GPUs/NPUs, LoRA, and GRPO;
- Covered activation offload, Megatron `expert_fc1` offload, and FSDP2 parameter CPU offload;
- Covered rollout, forward, backward, gradient clipping, optimizer step, and checkpoint paths;
- The fixed FSDP2 GRPO path successfully reached `global_step=1/1`.

## Scope

This PR modifies only the following 4 source files:

- `swift/callbacks/activation_cpu_offload.py`
- `swift/rlhf_trainers/grpo_trainer.py`
- `swift/rlhf_trainers/rollout_mixin.py`
- `swift/trainers/mixin.py`

Reproduction scripts, issue reports, and test files are not included in this PR.


swift/rlhf_trainers/grpo_trainer.py
```diff
@@ -2051,7 +2052,7 @@ def get_chunked_inputs(self, model_inputs, grpo_batch, start_idx, end_idx, origi
                 for ed in encoded_data:
                     ed.pop('_extra_kwargs', None)
                 chunk_model_inputs.update(
-                    to_device(template.data_collator(encoded_data, padding_to=current_length), self.model.device))
+                    to_device(template.data_collator(encoded_data, padding_to=current_length), self.accelerator.device))
```
hjh0119, line 2054: Could you also fix this on the GKDTrainer side?

hjh0119 (approved):

swift/rlhf_trainers/grpo_trainer.py
addsubmuldiv: ok, fixed

PR #9849, opened 2026-08-04
fix(infer): propagate streaming errors

RerankerGuo:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Background

`InferEngine.async_iter_to_iter()` consumes an asynchronous stream in a
background thread and forwards items through a synchronous queue.

When the asynchronous stream failed in strict mode, the producer thread
re-raised the exception before placing any terminal value in the queue.
The consumer remained blocked on `queue.get()` and could not receive the
original error. In non-strict mode the exception was queued, but the
consumer treated it as a normal end-of-stream marker.

## Changes

- Always place asynchronous stream exceptions in the consumer queue.
- Re-raise the queued exception from the caller thread when
  `strict=True`.
- Preserve the existing non-strict behavior, where the failed stream
  ends without raising.
- Add CPU regression tests for both strict and non-strict modes using a
  timeout queue so the pre-fix deadlock is safely detectable.

## Verification

- `PYTHONPATH=. .venv/bin/python -m unittest tests.infer.test_infer_engine`
- `PYTHONPATH=. .venv/bin/python tests/run.py --pattern test_infer_engine.py`
- `PYTHONPATH=. .venv/bin/python -m unittest tests.infer.test_infer_engine tests.general.test_data_preprocess.TestProviderMessagesPreprocess tests.general.test_data_preprocess.TestRejectedMessagesPreprocess`
- `.venv/bin/pre-commit run --all-files`
- `git diff --check`

## Impact

Only asynchronous streaming failures are affected. Successful stream
items and metrics handling are unchanged. Strict callers now receive the
original exception instead of waiting indefinitely; non-strict callers
continue to terminate the failed stream quietly.

## Experiment results

Before the fix, the strict-mode regression test timed out waiting for the
queue and raised `queue.Empty` while the producer thread printed the
original `RuntimeError`. After the fix, the caller receives that
`RuntimeError` directly, and both strict and non-strict tests pass.

### DIFF
--- a/swift/infer_engine/infer_engine.py
+++ b/swift/infer_engine/infer_engine.py
@@ -84,8 +84,6 @@ class InferEngine(BaseInferEngine, ProcessorMixin):
                 async for item in await async_iter:
                     queue.put(item)
             except Exception as e:
-                if getattr(self, 'strict', True):
-                    raise
                 queue.put(e)
             else:
                 queue.put(None)
@@ -103,6 +101,8 @@ class InferEngine(BaseInferEngine, ProcessorMixin):
             if output is None or isinstance(output, Exception):
                 prog_bar.update()
                 self._update_metrics(pre_output, metrics)
+                if isinstance(output, Exception) and getattr(self, 'strict', True):
+                    raise output
                 return
             pre_output = output
             yield output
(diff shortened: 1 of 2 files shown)


hjh0119 (approved):

hjh0119:
thanks!

PR #9852, opened 2026-08-04
fix(ci): use unique docker container names

RerankerGuo:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Background

The GPU CI launcher always names its Docker container `swift-ci-0` or
`swift-ci-1`. On a self-hosted runner, another active or stale workflow
can already own that global Docker name even after the GPU lock for the
current job is available.

This caused [run 30896752033](https://github.com/modelscope/ms-swift/actions/runs/30896752033)
to fail before tests started:

```text
Conflict. The container name "/swift-ci-0" is already in use
```

## Changes

- Include `GITHUB_RUN_ID`, `GITHUB_RUN_ATTEMPT`, and the selected GPU
  slot in each CI container name.
- Use `local-<pid>` and attempt `0` as deterministic local fallbacks.
- Add a CPU regression test that runs the launcher with stubbed
  `docker` and `flock` commands and verifies the complete `docker run`
  invocation.

## Verification

- `PYTHONPATH=. .venv/bin/python -m unittest tests.general.test_dockerci`
- `PYTHONPATH=. .venv/bin/python tests/run.py --pattern test_dockerci.py`
- `bash -n .dev_scripts/dockerci.sh`
- `.venv/bin/pre-commit run --all-files`
- `git diff --check`

## Impact

GPU selection and lock files are unchanged. Only the Docker container
name changes, preventing unrelated workflow runs from colliding in the
daemon-wide name namespace.

## Experiment results

Before the fix, the stubbed launcher emitted:

```text
--name swift-ci-0
```

With `GITHUB_RUN_ID=12345` and `GITHUB_RUN_ATTEMPT=2`, it now emits:

```text
--name swift-ci-12345-2-0
```

### DIFF
--- a/.dev_scripts/dockerci.sh
+++ b/.dev_scripts/dockerci.sh
@@ -14,6 +14,8 @@ PR_CHANGED_FILES="${PR_CHANGED_FILES:-}"
 echo "PR modified files: $PR_CHANGED_FILES"
 PR_CHANGED_FILES=${PR_CHANGED_FILES//[ ]/#}
 echo "PR_CHANGED_FILES: $PR_CHANGED_FILES"
+CI_RUN_ID=${GITHUB_RUN_ID:-local-$$}
+CI_RUN_ATTEMPT=${GITHUB_RUN_ATTEMPT:-0}
 idx=0
 for gpu in $gpus
 do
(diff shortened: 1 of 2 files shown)


RerankerGuo:
CI note: the failing CUDA unittest is unrelated to this PR. `test_container_name_includes_workflow_identity` passes in that run; the only error is `ModuleNotFoundError: transformers.models.gemma3` in `test_gemma3_template`. Lint and the other unittest job both pass. The repository-wide resolver regression is addressed by #9862, which keeps the Transformers 4.x pin through `pip install .`.

RerankerGuo:
Closing this to reduce the active review queue to the agreed maintenance capacity. The CI container-name fix is preserved, but it changes self-hosted runner infrastructure without maintainer feedback, so I am deferring it behind user-facing correctness fixes. The implementation and regression test remain at `6bca60d2449e3d3f4154e3f3e70049b5d7c6b3f5` for reopening or salvage.

Issue #9855, opened 2026-08-04
GRPO vLLM async rollout may ignore updated LoRA adapter in server mode


Zuozhuo:
### Checklist / 检查清单

- [x] I have searched existing issues, and this is a new bug report. / 我已经搜索过现有的 issues，确认这是一个新的 bug report。

### Bug Description / Bug 描述

In `ms-swift==4.3.1`, GRPO with vLLM server mode may not use the updated LoRA adapter during async rollout.

The issue appears under the following setting:

- `--use_vllm true`
- `--vllm_mode server`
- rollout server with `--vllm_use_async_engine true`
- rollout server with `--vllm_enable_lora true`
- GRPO + LoRA training
- multi-turn scheduler

The trainer can successfully send LoRA weight updates to the rollout server, but async rollout generation requests appear to keep sampling from the base model instead of the current LoRA policy. This disconnects policy update and rollout sampling, causing reward to remain noisy / non-convergent.

We confirmed this with a simple diagnostic reward: reward is higher when the generated content is longer. With native generate rollout, the model learns to generate longer outputs and reward increases clearly. With swift vLLM server async rollout, reward does not increase.

### How to Reproduce / 如何复现

Use a simple GRPO + LoRA setup with a toy length-based reward:

1. Train with native generate rollout:

   - single-turn prompt
   - reward is monotonically higher when generated string length is longer
   - result: reward increases as expected

2. Train with swift rollout server:

   - same model
   - same dataset
   - same reward
   - same GRPO settings
   - only change rollout backend to vLLM server async rollout

   Example rollout server flags:

   ```bash
   swift rollout \
     --model "$MODEL_PATH" \
     --vllm_enable_lora true \
     --vllm_use_async_engine true \
     --multi_turn_scheduler <custom_multi_turn_scheduler>
   ```

   Example trainer flags:

   ```bash
   swift rlhf \
     --rlhf_type grpo \
     --use_vllm true \
     --vllm_mode server \
     --vllm_server_base_url "$ROLLOUT_URL" \
     --multi_turn_scheduler <custom_multi_turn_scheduler>
   ```

   Observed result: reward stays noisy and does not increase, even for the trivial length-based reward.

During the vLLM server run, rollout server logs show successful LoRA update requests such as `update_adapter_flattened_param`, and trainer-side KL increases, which suggests the trainer-side LoRA parameters are changing. However, rollout behavior does not reflect the updated policy.

### Additional Information / 补充信息

Suspected root cause:

[...]

PR #9856, opened 2026-08-04
Fix missing LoRARequest injection in GRPOVllmEngine.infer_async

Zuozhuo:
## What does this PR do?

This PR fixes missing LoRARequest injection in `GRPOVllmEngine.infer_async()`.

Before this patch, `GRPOVllmEngine.infer()` injected the current GRPO LoRA adapter when it had been loaded, but the async single-request path did not have the same logic. Since multi-turn rollout uses `infer_engine.infer_async()` directly, async vLLM server rollout could generate without `lora_request` and sample from the base model instead of the updated LoRA policy.

This PR:
- extracts the GRPO LoRARequest construction into helper methods
- reuses the logic in both sync `infer()` and async `infer_async()`
- keeps explicitly provided `adapter_request` unchanged
- handles `engine.list_loras()` as either sync or awaitable for compatibility

## Validation

I validated the fix with a toy GRPO diagnostic reward where longer generated content receives higher reward.

- native generate rollout: reward increases as expected
- vLLM server async rollout before this patch: reward stays noisy and does not increase
- vLLM server async rollout after this patch: reward increases as expected

The rollout server also received LoRA update requests during training, so the issue was not weight update delivery but missing LoRARequest usage during async generation.

### DIFF
--- a/swift/infer_engine/grpo_vllm_engine.py
+++ b/swift/infer_engine/grpo_vllm_engine.py
@@ -1,4 +1,5 @@
 # Copyright (c) ModelScope Contributors. All rights reserved.
+import inspect
 import os
 import torch
 from PIL import Image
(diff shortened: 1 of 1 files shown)


Zuozhuo:
Related issue: #9855

hjh0119 (approved):

hjh0119:
thanks for your contribution!

PR #9859, opened 2026-08-05
fix(sampling): preserve remainder samples across shards

RerankerGuo:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates
- [ ] More Models or Datasets Support

# PR information

`SwiftSampling._get_dataset` calculated one fixed shard size with integer
division. When the dataset size was not divisible by the number of
`data_range` shards, every shard used that truncated size and the remainder
samples were never processed.

For example, 10 samples split across 3 processes produced shard sizes
`[3, 3, 3]`, silently dropping sample 9.

This PR distributes the remainder across the first shards. The resulting
shards remain contiguous and ordered, differ in size by at most one, and cover
each input exactly once. Datasets smaller than the shard count are also
handled, with empty shards only after all samples have been assigned.

The change is limited to dataset partitioning for `data_range`; sampling,
resume, and output behavior are unchanged.

## Experiment results

Before the fix:

```text
10 samples / 3 shards -> [3, 3, 3], sample 9 missing
FAILED (Runs=1, success=0, failures=1)
```

After the fix:

```text
10 samples / 3 shards -> [4, 3, 3], complete ordered coverage
 2 samples / 3 shards -> [1, 1, 0], complete ordered coverage

.venv/bin/python tests/run.py --test_dir tests/sample --pattern test_sampling.py
SUCCESS (Runs=1, success=1)

.venv/bin/pre-commit run --all-files
All hooks passed
```

PR #9850, opened 2026-08-04
fix(internvl): pass max patch count to vllm

RerankerGuo:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates
- [ ] More Models or Datasets Support

# PR information

Fixes #7631.

## Background

InternVL templates read the `MAX_NUM` environment variable into
`template.max_num`. The Transformers backend uses this value during
local image preprocessing, but the vLLM branch only emitted the image
placeholder and did not forward the patch limit.

As a result, vLLM used the model configuration's default
`max_dynamic_patch`, potentially producing prompts that exceeded the
model context even when users set `MAX_NUM`.

vLLM's InternVL processor accepts `max_dynamic_patch` through its
multimodal processor kwargs, and `VllmEngine` already forwards
`inputs.mm_processor_kwargs` with each request.

## Changes

- Forward `template.max_num` as `max_dynamic_patch` for InternVL
  templates in vLLM mode.
- Use `setdefault` so an explicit per-request `max_dynamic_patch`
  remains authoritative.
- Add CPU tests for default propagation, explicit override precedence,
  and unchanged Transformers behavior.

## Verification

- `PYTHONPATH=. .venv/bin/python -m unittest tests.general.test_internvl_template`
- `PYTHONPATH=. .venv/bin/python tests/run.py --pattern test_internvl_template.py`
- `PYTHONPATH=. .venv/bin/python -m unittest tests.general.test_internvl_template tests.general.test_data_preprocess.TestProviderMessagesPreprocess tests.general.test_data_preprocess.TestRejectedMessagesPreprocess`
- `.venv/bin/pre-commit run --all-files`
- `git diff --check`

## Impact

The change only adds a vLLM multimodal processor kwarg for InternVL
templates. Transformers preprocessing is unchanged. Existing explicit
per-request values are preserved.

## Experiment results

Before the fix, the vLLM regression test observed an empty
`mm_processor_kwargs` dictionary despite `max_num=6`. After the fix it
contains `{"max_dynamic_patch": 6}`; all three template tests and ten
related tests pass.

PR #9862, opened 2026-08-05
fix(ci): keep the transformers<5.0 pin through the package install

he-yufeng:
The pin line in the CUDA path runs before `pip install .`, but pip keeps no constraints across invocations. The package install then re-resolves trl (floor `transformers>=4.56.2`) against framework.txt's `<5.15.0` and jumps to transformers 5.14.1, after which the suite dies importing `transformers.models.gemma3`. Every PR is red on this right now (confirmed on mine and on #9854, same error in the same step).

Carrying the pin into the package install keeps the resolver on the 4.x line the suite is written for. Verified with the actual resolver: `trl==0.29.1` + `transformers<5.0` in one run selects transformers 4.57.6, which satisfies trl's floor and keeps gemma3 importable.

The NPU path already carries the pin inside its own command and is untouched. Whether framework.txt itself should drop `<5.15.0` for users is a bigger support-range question I left alone here.

### DIFF
--- a/.dev_scripts/ci_container_test.sh
+++ b/.dev_scripts/ci_container_test.sh
@@ -228,8 +228,12 @@ if [ "$MODELSCOPE_SDK_DEBUG" == "True" ]; then
         pip uninstall autoawq -y
         pip install optimum
         pip install diffusers
+        # The pin must ride into the package install itself: a bare
+        # `pip install .` re-resolves trl (>=4.56.2 floor) against
+        # framework.txt's <5.15.0 and jumps to transformers 5.x, whose
+        # layout the test suite does not support yet.
         pip install "transformers<5.0" "peft<0.19"
-        pip install .
+        pip install "transformers<5.0" .
         pip install auto_gptq bitsandbytes deepspeed -U -i https://mirrors.aliyun.com/pypi/simple/
     fi
 else
--- a/tests/run.py
+++ b/tests/run.py
@@ -559,16 +559,23 @@ def print_table_result(df):
 
 def main(args):
     runner = TimeCostTextTestRunner()
+    # Snapshot the dep versions this process starts with. A suite can
+    # pip-install its model's own pinned requirements into the shared env
+    # (modelscope remote code), which would otherwise poison every later
+    # suite in the same chunk.
+    record_env_versions()
     if args.suites is not None and len(args.suites) > 0:
         logger.info('Running: %s' % ' '.join(args.suites))
         test_suite = gather_test_suites_in_files(args.test_dir, args.suites, args.list_tests)
     else:
         test_suite = gather_test_cases(os.path.abspath(args.test_dir), args.pattern, args.list_tests)
     if not args.list_tests:
-        result = runner.run(test_suite)
+        result_list = []
+        for sub_suite in test_suite:
+            result_list.extend(collect_test_results(runner.run(sub_suite)))
+            restore_env_versions()
         logger.info('Running case completed, pid: %s, suites: %s' % (os.getpid(), args.suites))
-        result = collect_test_results(result)
-        df = test_cases_result_to_df(result)
+        df = test_cases_result_to_df(result_list)
         if args.result_dir is not None:
             save_test_result(df, args)
         else:
--- a/tests/run.py
+++ b/tests/run.py
@@ -14,6 +14,7 @@ import torch
 import unittest
 import yaml
 from fnmatch import fnmatch
+from importlib import metadata
 from model_tag import ModelTag, commit_model_ut_result
 from pathlib import Path
 from test_utils import get_case_model_info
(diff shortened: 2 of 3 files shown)


RerankerGuo:
I traced the full failing run and found a second resolver mutation that this patch does not cover. In run 31179195838, the initial pin correctly installs Transformers 4.57.6. Later, `tests/tuners/test_merged_linear.py` loads `damo/nlp_structbert_sentence-similarity_chinese-base` with remote code; that model installs its own requirements (`huggingface-hub==0.25.*`, `transformers==4.48.3`) into the shared test environment. The log then shows `Successfully installed ... transformers-4.48.3`, and only afterward `test_gemma3_template.py` fails with `ModuleNotFoundError: transformers.models.gemma3`.

The same sequence appears in the five currently red PR runs (#9850, #9852, #9859, #9870, #9871). Therefore carrying `transformers<5.0` into `pip install .` is useful for the initial install but will not make these runs green. The fix also needs to prevent/isolate the remote StructBERT requirements install, or restore the supported dependency set after that suite. Merely moving the Gemma import would mask the shared-environment corruption.

he-yufeng:
@RerankerGuo Good catch, that StructBERT install explains the rest of the red runs. I went with the restore option in 5d534f7.

Preventing the model's requirements install would break test_merged_linear itself, since the remote code needs its pinned set to load. Isolating every suite in its own venv would work but is a much bigger change to the runner. Restoring after the suite keeps both sides intact: StructBERT runs on 4.48.3 inside its own suite, and the next suite starts on the supported set again.

What the patch does in tests/run.py:

- `record_env_versions()` snapshots the installed transformers/peft/huggingface-hub versions right after each env's requirements and dependencies are installed, i.e. the state the suites are written for (4.57.6 in the current image).
- `restore_env_versions()` runs when a suite finishes and reinstalls only what drifted. It is a no-op for suites that leave the environment alone.
- The hook sits in `wait_for_free_worker` where a completed worker is reaped, plus after each suite in the serial `run_case_in_env` path. With the CI default `--parallel 1` the next suite only starts after the previous one is reaped, so the restore always lands between the polluting suite and the next one.

Your traced sequence is covered: test_merged_linear gets its 4.48.3 during its own run, then the restore puts 4.57.6 back before test_gemma3_template starts. huggingface-hub is in the snapshot list too, though reinstalling transformers 4.57.6 would pull it back above the required floor either way.

Verified locally with py_compile plus a standalone run of the record/restore logic (records versions, no-op without drift, reinstalls the pinned version with drift). I cannot reproduce the modelscope remote-code install on this machine, so the real check is this PR's CI run.

he-yufeng:
Follow-up on the run of the restore commit: it still went red, and the log shows exactly why. The two errors are `test_dataset` failing to import (`AssistantVocabTranslatorCache` / `QuantizedCacheConfig` missing from the transformers on disk at that moment), i.e. it imported while the StructBERT 4.48.3 install was active.

The gap was in my first patch's coverage, not in the approach. With the CI default `--parallel 1`, all remaining suites of an env go into a single chunk that runs in one process (`--suites <all of them>`), so the pollution lands mid-process and the worker-reap restore between processes never gets a chance to run. d7c910dc closes that: `main()` now runs each per-file sub-suite in turn and restores the pinned versions between them, which covers the chunk interior. Single-suite workers are untouched.

Verified locally: py_compile plus a unittest-semantics check that the gathered suite iterates per-file. The CI run on this head is the real proof again.

he-yufeng:
The run on d7c910 still failed the same way, and the log pinned the race: the restore's `pip install transformers==4.57.6` takes ~40s with the package fully uninstalled mid-flight, and this runner uses `--parallel 2`, so the pip window overlapped the sibling worker's imports (the failures land exactly on the uninstall timestamps). Restoring between suites only serializes one side of the pool.

d25384b closes it on the mutation side rather than the restore side:

- `tests/tuners/test_merged_linear.py` (the StructBERT remote-code suite that pip-installs the model's pinned requirements) is now on the `isolated` list, and isolated suites get a barrier on both sides when parallel > 1: the pool drains, the suite runs alone, the pins go back, then the pool resumes.
- The reap-point restore now only fires when no sibling worker is mid-run, and there's a restore after the final drain so env handoff between envs stays clean. With parallel == 1 the behavior is unchanged (the pool is trivially idle at reap).

Chunk suites are still assumed not to pip-install; anything else that does should join the isolated list. Verified locally as before (compile plus flow-shape check); this CI run is the proof.

he-yufeng:
Status on d25384b: the unittest lane now passes with the barrier fix (the parallel-2 race is covered). The lint red is inherited from main, not this branch: main's own lint run today (32009097888) fails the same double-quote-string-fixer hook with "files were modified by this hook", and running the fixer over my changed files (tests/run.py, tests/run_config.yaml) produces no edits. Once main's lint settles this should go green as-is.

he-yufeng:
Correction on the lint note above: I said the red was inherited from main. It is not reproducible at all — `pre-commit run --all-files` (the exact CI command) passes on this head locally, and the double-quote hook makes no edits anywhere in the repo. The CI log shows "files were modified by this hook" without naming one, which with a clean local run points at a stale pre-commit cache on the runner. A re-run of the lint job should clear it.

hjh0119:
Sorry for the late review. The root cause turned out to be a few stale unit tests loading legacy modelscope remote-code models (StructBERT/CoROM), which pip-install their own pinned dependencies (e.g. transformers==4.48.3) into the shared CI environment — exactly the pollution @RerankerGuo traced, thanks for the investigation!

We're removing those tests entirely, and also dropping the transformers<5.0 / peft<0.19 pins in the CI container so it installs the latest versions allowed by requirements/framework.txt — swift is now developed against transformers 5.x. With the pollution source gone, the env-restore/isolation machinery in this PR is no longer needed, so I think we can close this in favor of the simpler fix. #9938

he-yufeng:
Makes sense. With the StructBERT/CoROM suites gone there is nothing left for the restore and barrier machinery to guard, and the pin would just fight the 5.x move. Closing in favor of #9938. Thanks for digging into the root cause.

PR #9866, opened 2026-08-06
[Infer] Fix transformers streaming logprobs

taking-lying-flat:
## Summary

- retain the `LogitsStreamer` instance used by PT/Transformers generation so streaming responses can consume its logits

## Root cause

The PT streaming path constructed `LogitsStreamer()` inline inside `LogitsProcessorList`, while the local `logits_streamer` variable remained `None`. The collection path therefore returned early for every streamed chunk, and `logprobs` was serialized as `null`.

## Impact

Requests using the Transformers backend with both `stream=true` and `logprobs=true` now return populated logprobs. Non-streaming inference and streaming requests without logprobs are unchanged.

## Validation

- `python -m py_compile swift/infer_engine/transformers_engine.py`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/swift/infer_engine/transformers_engine.py
+++ b/swift/infer_engine/transformers_engine.py
@@ -255,7 +255,8 @@ class TransformersEngine(InferEngine):
 
         logits_streamer = None
         if generation_config.output_logits:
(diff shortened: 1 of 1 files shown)


hjh0119 (approved):

hjh0119:
thanks

PR #9873, opened 2026-08-08
Optimize logprob computation in Transformers inference and GRPO

taking-lying-flat:
## Summary

- Vectorize `TransformersEngine` logprob post-processing with batched `topk`, gather, and host transfer operations.
- Skip GRPO old-policy logprob forwards when the rollout lifecycle is aligned with optimizer updates.
- Preserve eager old-policy logprobs only for pre-forward consumers: reward KL, teacher/OPSD paths, and enabled Liger rollout correction.
- Avoid toggling gradient checkpointing when no student inference forward is required.

## Motivation

The Transformers rollout path previously sorted the full vocabulary and synchronized individual token values to the host. GRPO also computed old-policy logprobs unconditionally, even when the policy could not change during the rollout lifecycle and the loss can reuse detached current-policy logprobs.

The new GRPO gate uses the full generation interval:

```text
steps_per_generation * sequence_parallel_size * num_iterations
```

and keeps the extra forward only when the interval can cross an optimizer update or another consumer needs old logprobs before the training forward.

## Impact

On a controlled RTX A1000 BF16 benchmark with a 33.6M-parameter Qwen2 model, batch size 4 and sequence length 512, removing the redundant old-policy forward improved effective policy-training throughput by:

- 32.5% for `num_iterations=1`
- 16.1% for `num_iterations=2`
- 8.0% for `num_iterations=4`

The improvement decreases as expected when the single old-policy forward is amortized over more training iterations.

## Validation

- `pytest -q tests/infer/test_transformers_engine_logprobs_unit.py`
- Local GRPO old-policy gate matrix: 5 tests and 5 subtests passed
- Teacher/OPSD routing suite: 27 tests passed
- `git diff --check`

### DIFF
--- a/swift/infer_engine/transformers_engine.py
+++ b/swift/infer_engine/transformers_engine.py
@@ -206,21 +206,38 @@ class TransformersEngine(InferEngine):
     @staticmethod
     def preprocess_logits(batched_logits: Optional[List[torch.Tensor]], batched_generate_ids: torch.Tensor,
                           top_logprobs: Optional[int]):
-        top_logprobs = top_logprobs or 1
         batch_size = batched_generate_ids.shape[0]
         if batched_logits is None:
             return None
-        batched_logprobs = []
-        for i in range(batch_size):
-            logprobs_list = []
-            generate_ids = batched_generate_ids[i]
-            for j, logits in enumerate(batched_logits):
-                token = generate_ids[j].item()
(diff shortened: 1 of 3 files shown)


taking-lying-flat:
Closing because this branch unintentionally included an unrelated Transformers logprobs commit. Replaced by a clean GRPO-only branch.

PR #9589, opened 2026-06-17
Fix NPU Qwen3.5 grpo bugs

addsubmuldiv:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates
- [x] More Models or Datasets Support

# PR information

This PR enables **Qwen3.5 / Qwen3.5-MoE** GRPO on **NPU**, covering both **colocate** and **server** vLLM rollout paths, and documents the known limitation of **MoE fused expert LoRA** under **PEFT + Transformers 5**.

The changes are grouped into four areas:

```text
1. Qwen3.5 colocate / basic compatibility
2. vLLM-Ascend MoE weight-sync layout compatibility
3. vLLM server-mode weight synchronization compatibility
4. Documentation for PEFT MoE fused expert LoRA limitations
```

This PR only contains framework code changes and does not include local smoke scripts.

## 1. Qwen3.5 RoPE config compatibility with vLLM 0.18 + Transformers 5

Files involved:

- `swift/infer_engine/vllm_engine.py`

### Problem

Qwen3.5 depends on the config / RoPE parameter handling logic in Transformers 5. In the Qwen3.5 config path of vLLM 0.18.x, `ignore_keys_at_rope_validation` may still be passed in as a `list`; however, the Transformers 5 side treats it as a `set` and performs set operations on it, which leads to a type error.

### Fix

When creating the vLLM engine, this PR temporarily patches `PretrainedConfig.convert_rope_params_to_dict()` so that it only converts `ignore_keys_at_rope_validation` from `list` to `set` when needed.

### Upstream status

- This is a compatibility issue specific to the combination of **vLLM 0.18.x** and **Transformers 5**.
- Later vLLM releases have adjusted the related implementation in the Qwen3.5 config path.
- References:
  - https://github.com/vllm-project/vllm/issues/36411
  - https://github.com/vllm-project/vllm/pull/37338

## 2. MoE weight synchronization layout compatibility for vLLM-Ascend

Files involved:

- `swift/model/npu_patch/vllm_ascend.py`
- `swift/model/npu_patch/vllm_ascend_moe.py`
- `swift/rlhf_trainers/rollout_mixin.py`
- `swift/rlhf_trainers/utils.py`

### Problem

During GRPO training, training-side weights need to be synchronized to the vLLM rollout side. For Qwen MoE on vLLM-Ascend, MoE expert weights have backend-specific layouts:

- training-side weights may come from HF / FSDP2 / Megatron layouts;
[...]

PR #9877, opened 2026-08-10
fix(grpo): make revert_runtime_names_to_checkpoint idempotent to avoid double-expanding prefix renamings during vLLM weight sync

loveleaves:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## What does this PR do?

Fixes a crash in GRPO online weight sync (colocated vLLM) for models whose
transformers>=5 `WeightRenaming` rule is a **prefix substitution** — most
notably **Qwen3.5-MoE** (`qwen3_5_moe`).

`revert_runtime_names_to_checkpoint()` applies the *reversed* renaming rule
unconditionally. For prefix-anchored renames the reversed rule also matches
names that are **already in checkpoint form**, re-expanding the prefix a second
time and corrupting every parameter name. The corrupted names then fail to load
into vLLM.

GRPO training of Qwen3.5-MoE currently crashes on the first rollout weight sync
(before any optimization step), for both LoRA and full tuning:

```
KeyError: 'language_model.language_model.layers.0.mlp.experts.w13_weight'
```

## Root cause

`_build_reverse_renamings(model)` inverts each transformers `WeightRenaming` via
`reverse_transform()`. For Qwen3.5-MoE the forward rule is:

```
source: ^model.language_model   ->   target: model
```

so the reversed rule becomes:

```
^model   ->   model.language_model
```

At weight-sync time the runtime `state_dict` already carries checkpoint-form
names such as `model.language_model.layers.0.mlp.experts.gate_up_proj`. Applying
the reversed rule `^model -> model.language_model` to a name that *already*
begins with `model.language_model` matches the leading `model` and re-expands it:

```
model.language_model.layers.0...   ->   model.language_model.language_model.layers.0...
```

After vLLM's `hf_to_vllm_mapper` (`model.language_model. -> language_model.model.`)
this reaches the vLLM model as `language_model.language_model...experts.w13_weight`,
and the fused-expert loader's `params_dict[name]` lookup raises `KeyError`.

The current loop applies the reverse rule blindly, with no protection against
over-matching:

```python
new_state_dict = {}
for name, param in state_dict.items():
    try:
        new_name, _ = rename_source_key(name, reverse_renamings, [])
    except Exception:
        new_name = name
    new_state_dict[new_name] = param
return new_state_dict
```

This regression was introduced together with the function in #9572
("fix gemma4-12b weight sync"). It is correct for gemma4's 1:1 rename
(`model.vision_embedder ↔ model.embed_vision`, which is idempotent under the
reverse rule) but breaks on prefix-anchored renames.

## The fix

A legitimate runtime→checkpoint revert is **idempotent**: applying the reverse
rules to a name that is already in checkpoint form must be a no-op. We detect
over-matching by re-applying the reverse rules once — if the name keeps changing,
it was already a checkpoint name and we keep the original.

```python
new_state_dict = {}
for name, param in state_dict.items():
    try:
        new_name, _ = rename_source_key(name, reverse_renamings, [])
        # Guard against over-matching reverse rules. Prefix-anchored renamings
        # such as ``^model.language_model <-> model`` invert to
        # ``^model -> model.language_model``; that reverse rule ALSO matches
        # runtime names that are already in checkpoint form (e.g. Qwen3.5-MoE
        # ``model.language_model.layers...``), re-expanding the prefix into
        # ``model.language_model.language_model...`` and corrupting every key.
        # A genuine runtime->checkpoint revert is idempotent: re-applying the
        # reverse rules must not change the name again. If it keeps growing,
        # the name was already a checkpoint name and must be left untouched,
        # otherwise vLLM weight sync fails with KeyError on e.g.
        # 'language_model.language_model...experts.w13_weight'.
        if new_name != name:
            reapplied, _ = rename_source_key(new_name, reverse_renamings, [])
            if reapplied != new_name:
                new_name = name
    except Exception:
        new_name = name
    new_state_dict[new_name] = param
return new_state_dict
```

The change is fully localized to `revert_runtime_names_to_checkpoint()` and does
not touch the reverse-rule construction, the vLLM mapper, or the MoE weight
loaders.

## Affected file

- `swift/rlhf_trainers/utils.py` — `revert_runtime_names_to_checkpoint()`

## How was it tested?

- **Unit**: applying the reverse rule to `model.language_model.*` names is now a
  no-op (idempotent), whereas before it produced `model.language_model.language_model.*`.
- **End-to-end (CPU / meta device, no accelerator required)**: the full pipeline
  `revert_runtime_names_to_checkpoint → hf_to_vllm_mapper → AutoWeightsLoader →
  fused-expert lookup` loads every Qwen3.5-MoE weight with no `KeyError`.
  Before the fix the same pipeline raises
  `KeyError: language_model.language_model...w13_weight`.
- **Regression**: gemma4-style 1:1 renames (`embed_vision ↔ vision_embedder`) are
  unchanged — the guard only reverts names that would otherwise be over-expanded,
  and idempotent renames never trigger it.
- Repro command: `swift rlhf --rlhf_type grpo --model_type qwen3_5_moe
  --vllm_mode colocate ...` (both LoRA and full fail identically without the fix,
  confirming the bug is tuner-independent and lives in the shared weight-sync path).

## Related work (non-overlapping)

- #9572 (merged) — introduced `revert_runtime_names_to_checkpoint` for gemma4;
  did not consider prefix-anchored renames.
- #9589, #9800 (merged) — fix NPU Qwen3.5 GRPO issues in the MoE expert weight
  loaders / vLLM-Ascend patches; **neither modifies the revert loop**, so this
  bug remains present on `main`.

## Checklist

- [x] Bug reproduced on `main`
- [x] Fix is minimal and localized to `revert_runtime_names_to_checkpoint`
- [x] No behavior change for models with idempotent (1:1) renames
- [x] Fixes Qwen3.5-MoE / prefix-anchored renamed models under GRPO weight sync

### DIFF
--- a/swift/rlhf_trainers/utils.py
+++ b/swift/rlhf_trainers/utils.py
@@ -1191,6 +1191,21 @@ def revert_runtime_names_to_checkpoint(model, state_dict):
     for name, param in state_dict.items():
         try:
             new_name, _ = rename_source_key(name, reverse_renamings, [])
+            # Guard against over-matching reverse rules. Prefix-anchored
+            # renamings such as ``^model.language_model <-> model`` invert to
+            # ``^model -> model.language_model``; that reverse rule ALSO matches
+            # runtime names that are already in checkpoint form (e.g. Qwen3.5-MoE
+            # ``model.language_model.layers...``), re-expanding the prefix into
+            # ``model.language_model.language_model...`` and corrupting every key.
+            # A genuine runtime->checkpoint revert is idempotent: re-applying the
+            # reverse rules must not change the name again. If it keeps growing,
+            # the name was already a checkpoint name and must be left untouched,
+            # otherwise vLLM weight sync fails with KeyError on e.g.
+            # 'language_model.language_model...experts.w13_weight'.
+            if new_name != name:
+                reapplied, _ = rename_source_key(new_name, reverse_renamings, [])
+                if reapplied != new_name:
+                    new_name = name
         except Exception:
             new_name = name
         new_state_dict[new_name] = param


hjh0119:
let me check it

hjh0119:
can't reproduce with vllm 0.25.1 transformers 5.12.0, can you update and retry?

loveleaves:
> can't reproduce with vllm 0.25.1 transformers 5.12.0, can you update and retry?

Thanks for checking. I dug into why it reproduces for me but not for you — it's a **transformers-version issue, not a vllm one**, and the version difference is exactly what hides it.

The corruption happens in swift's `revert_runtime_names_to_checkpoint`, **before** anything reaches vLLM. That function applies the *reverse* of transformers' `WeightRenaming` rules, and the reverse rule itself changed between transformers versions.

I verified this by running the real `get_model_conversion_mapping` API on both versions (CPU-only, meta device, no XPU):

**transformers 5.12.0 (your env) — reverse rule is idempotent, no bug:**
```
fwd    : ^model\.language_model\.(.+)$              -model.\1
reverse: ^model\.(?:(?!language_model\.))(.+)$      -model.language_model.\1
runtime: model.language_model.layers.0.mlp.experts.gate_up_proj
revert : model.language_model.layers.0.mlp.experts.gate_up_proj   # no-op
```

**transformers 5.5.1 (my env, pinned by our downstream vllm_kunlun stack) — reverse rule over-matches:**
```
fwd    : model.language_model  -^model            # no anchor, no capture group
reverse: ^model                -model.language_model   # bare prefix, matches ANY 'model...'
runtime: model.language_model.layers.0.mlp.experts.gate_up_proj
revert : model.language_model.language_model.language_model.layers.0.mlp.experts.gate_up_proj   # double prefix 
```

The `5.5.1` reverse rule is a bare `^model -model.language_model`, so it re-expands runtime names that are **already** in checkpoint form. After vLLM's `hf_to_vllm_mapper` (`model.language_model. -language_model.model.`) this becomes `language_model.language_model...experts.w13_weight`, and the MoE fused-expert loader (`load_fused_expert_weights`, `params_dict[name]` with no guard) raises `KeyError`.

Two important consequences:

1. **The real fix landed in transformers itself** — 5.12.0 added a negative lookahead `(?!language_model\.)` that makes the reverse transform idempotent. That's why you can't reproduce on 5.12.0 regardless of vllm version; the corruption never happens upstream of vLLM.
2. **Dense vs MoE also matters at the vLLM layer.** Non-expert weights hit vLLM's guarded path (`if name not in params_dict: warning; continue`, silently skipped), while the MoE fused-expert loader does an unguarded `params_dict[name]` → hard `KeyError`. So a dense model may *look* fine while its weight sync is silently being dropped; only MoE surfaces it as a crash. If you tested with a dense/text model that would also explain the non-repro.

So the guard in this PR reproduces, at the swift layer, the same idempotency guarantee transformers only introduced in 5.12.0. It's a **no-op on 5.12.0** (the reverse rule is already idempotent, so the guard never triggers) and it fixes the crash for anyone pinned to older transformers (5.5.1 here).

loveleaves:
Below is my minimal reproduction script.
```python
import torch, transformers
from transformers.core_model_loading import WeightRenaming, rename_source_key
from transformers.conversion_mapping import get_model_conversion_mapping
from transformers.models.qwen3_5_moe.modeling_qwen3_5_moe import Qwen3_5MoeForConditionalGeneration
from transformers.models.qwen3_5_moe.configuration_qwen3_5_moe import Qwen3_5MoeConfig

cfg = Qwen3_5MoeConfig()
cfg.text_config.num_hidden_layers = 1
cfg.text_config.num_experts = 2
with torch.device("meta"):
    m = Qwen3_5MoeForConditionalGeneration(cfg)

conv = get_model_conversion_mapping(m, add_legacy=False)
ren = [c for c in conv if isinstance(c, WeightRenaming)]
rev = [c.reverse_transform() for c in ren[::-1]]

n = "model.language_model.layers.0.mlp.experts.gate_up_proj"
once, _ = rename_source_key(n, rev, [])
print("transformers", transformers.__version__)
print("runtime     :", n)
print("revert once :", once)
print(">>> over-match (bug):", once != n)
# 5.5.1  -> over-match=True  （double prefix, trigger KeyError）
# 5.12.0 -> over-match=False （no-op，no bug）
```

hjh0119:
Thanks for digging into this!

This looks like a defect in transformers itself rather than in swift. In v5.5 the qwen3_5_text rule was a bare renaming with no guard. Upstream fixed this in v5.6 by adding a negative lookahead (huggingface/transformers#45340, huggingface/transformers#45358).

Given that, I'd prefer to have users upgrade rather than carry a workaround in the weight-sync path. We'll also consider raising the minimum required transformers version so an older install fails fast with a clear message instead of surfacing as a KeyError here.

PR #9886, opened 2026-08-11
fix(template): keep MiMo dependencies optional

RerankerGuo:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

MiMo-V2.5 support added `qwen_vl_utils` as a module-level import in
`swift.template.templates.mimo`. Because all template modules are registered
when `swift.template` is imported, environments without that model-specific
optional dependency could no longer import unrelated template, dataset,
inference, trainer, or export modules.

This is visible in the NPU job for #9880, where 34 test modules failed during
collection with:

```text
ModuleNotFoundError: No module named 'qwen_vl_utils'
```

The dependency is already declared in the MiMo model's `requires` list. This
change follows the existing Qwen, Llava, Dots, and Valley template pattern by
importing `fetch_image` and `fetch_video` only when MiMo media is actually
processed. MiMo users still receive the dependency error at the relevant
feature boundary, while unrelated imports remain available.

A subprocess regression test explicitly blocks every `qwen_vl_utils` import
and verifies that `swift.template` still imports successfully.

## Experiment results

Before:

```text
.venv/bin/python -m unittest tests.general.test_optional_template_dependencies
ModuleNotFoundError: No module named 'qwen_vl_utils'
FAILED (errors=1)
```

After:

```text
.venv/bin/python tests/run.py \
  --test_dir tests/general \
  --pattern test_optional_template_dependencies.py
SUCCESS (Runs=1, success=1)

.venv/bin/python -m unittest tests.general.test_model
Ran 1 test
OK

.venv/bin/pre-commit run --all-files
All hooks passed
```

The local environment does not have `qwen-vl-utils` installed, so the direct
`swift.template`, `swift.infer_engine.protocol`, and `swift.dataset` imports
also verify the real missing-dependency path.

### DIFF
--- a/swift/template/templates/mimo.py
+++ b/swift/template/templates/mimo.py
@@ -1,7 +1,6 @@
 # Copyright (c) ModelScope Contributors. All rights reserved.
 import torch
 from dataclasses import dataclass, field
-from qwen_vl_utils import fetch_image, fetch_video
 from typing import Any, Dict, List, Optional
 
 from ..base import Template
@@ -32,6 +31,8 @@ class MiMoV2Template(Template):
     norm_bbox = 'none'
 
     def replace_tag(self, media_type, index, inputs: StdTemplateInputs) -> List[Context]:
+        from qwen_vl_utils import fetch_image, fetch_video
+
         assert media_type in {'image', 'video'}
         kwargs = {'image_patch_size': self.processor.image_processor.patch_size}
         if media_type == 'image':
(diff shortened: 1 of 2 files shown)


RerankerGuo:
CI note: the new optional-dependency regression test passes in the CUDA run
(`test_template_import_without_qwen_vl_utils`, 7.8s). The run's only error is
the repository-wide `ModuleNotFoundError: transformers.models.gemma3` in
`test_gemma3_template.py`, after the shared environment dependency mutation
already traced on #9862:
https://github.com/modelscope/ms-swift/pull/9862#issuecomment-5224213936.
Lint passes, and the NPU unittest is still running.

hjh0119 (approved):

PR #9921, opened 2026-08-15
[Megatron] Fix duplicated GRPO KL gathering

taking-lying-flat:
## Summary

- Slice rollout-group KL values back to the samples owned by the current rank before the global gather.
- Preserve the same global rank ordering used when rewards are gathered for advantage computation.
- Add regression coverage for equal and uneven per-rank rollout sample counts.

## Root cause

Megatron GRPO first gathers local samples across the rollout group (`TP x PP x CP`) and computes KL for that full concatenated sample set on every rank. `_compute_kl_from_batches` then gathered those already group-wide KL values globally.

Rewards follow a different path: each rank scores only its locally owned samples, then gathers those local rewards globally once. As a result, the KL tensor contained replicated rollout-group entries and could be larger than the rewards tensor. With the 4-way parallel configuration reported in #9917, this produced the deterministic `512` versus `2048` mismatch when applying the KL reward penalty.

## Fix

The rollout-group sample counts and rank identify the exact slice originally owned by each process. Only that local KL slice is passed to the existing global gather, so KL and rewards now use identical sample ownership and ordering. The implementation supports non-uniform sample counts within the rollout group and asserts that the concatenated KL tensor matches the gathered counts.

### DIFF
--- a/swift/megatron/trainers/grpo_trainer.py
+++ b/swift/megatron/trainers/grpo_trainer.py
@@ -13,7 +13,7 @@ from swift.infer_engine.protocol import RequestConfig, RolloutInferRequest, Roll
 from swift.megatron.arguments import MegatronArguments
 from swift.megatron.utils import RouterReplayHelper, get_padding_to, set_router_replay_data
 from swift.rl_core.advantage import (compute_advantages, compute_reward_metrics, compute_teacher_kl_per_token,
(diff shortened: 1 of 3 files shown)


hjh0119 (approved):

hjh0119:
thanks for the fix

PR #9923, opened 2026-08-15
fix: handle tool messages in last round loss scaling

Bakuma-sea:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Fix incorrect `last_round` loss scaling for agent conversations containing tool responses.

Previously, `get_last_user_round` only recognized messages with the `user` role. In tool-call conversations, a `tool` message starts the final query-response round, so the last round could be identified incorrectly and loss scaling could be applied to earlier assistant responses.

This change treats both `user` and `tool` messages as valid round-start messages. It also adds a regression test for a conversation ending with a tool response followed by the final assistant response.

## Experiment results

- Added regression test: `tests/loss_scale/test_last_user_round.py`
- `git diff --check` passed.

### DIFF
--- a/swift/template/utils.py
+++ b/swift/template/utils.py
@@ -166,17 +166,7 @@ def split_str_parts_by(text: str, delimiters: List[str], regex_mode: bool = Fals
 
 
 def get_last_user_round(messages):
-    """Get the index of the last round's user or tool message.
-
-    A round is a (query, assistant) pair; the query role is 'user' or 'tool'.
-    For agent / tool-call data the last round's query is usually a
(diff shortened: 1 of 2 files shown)


tastelikefeet (approved):

tastelikefeet:
Hi, can you lint the code via 'pre-commit run --all-files' please?

Bakuma-sea:
> Hi, can you lint the code via 'pre-commit run --all-files' please?

Thanks for the reminder. I ran `pre-commit run --all-files`, applied the yapf formatting changes, and all hooks pass now.

PR #9952, opened 2026-08-20
Fix colocated vLLM cleanup ordering on rollout failures

pureoxygen123:
## Summary

Fix the GPU memory transition order between colocated vLLM and the trainer when rollout generation fails.

Previously, vLLM was put to sleep only after rollout generation completed successfully. If synchronous, asynchronous, or multi-turn rollout raised an exception, the sleep call was skipped. Exiting `offload_context` would then reload the trainer model onto the GPU while vLLM was still occupying GPU memory, potentially causing an OOM and masking the original rollout error.

This change applies the fix to both the Transformers and Megatron rollout paths.

## Changes

- Run colocated vLLM cleanup in a `finally` block before `offload_context` reloads the trainer.
- Apply the same lifecycle ordering to the Megatron rollout path.
- Attempt every cleanup operation even if an earlier operation fails.
- Preserve the original rollout exception when cleanup also fails.
- Log every cleanup failure, including failures after the first one.
- Continue raising the first cleanup error when rollout itself succeeded.
- Add regression tests for successful rollouts, failed rollouts, and cleanup failures.

## Tests

`python -m pytest tests/utils/test_rollout_offload_order.py -v` passed all 3 tests.


hjh0119 (approved):

hjh0119:
Just curious, could you tell me under what circumstances rollout failures might occur?

PR #9964, opened 2026-08-21
fix(train): fail fast on incompatible DLRover flash checkpoint APIs

Excelius-Wang:
## Summary

- validate the DLRover Flash Checkpoint API before loading the model
- fail fast with an actionable source-install command when `blocking` or `max_steps` is unavailable
- add regression tests for the released DLRover 0.6.1 API and the current source API

## Motivation

The latest published DLRover release, 0.6.1, does not accept the `blocking` and `max_steps` arguments used by ms-swift. The incompatibility is currently discovered only while saving a checkpoint or shutting down training.

Silently dropping these arguments would be unsafe because `blocking=True` prevents the final checkpoint from being skipped while a previous asynchronous save holds the shared-memory lock. This change keeps the current checkpoint semantics and reports the incompatible dependency before model loading.

## Validation

- focused unit tests: 2 passed
- project test runner: 2/2 passed
- all relevant pre-commit hooks passed
- `git diff --check` passed

Related to #9912.

### DIFF
--- a/swift/trainers/mixin.py
+++ b/swift/trainers/mixin.py
@@ -64,24 +64,6 @@ logger = get_logger()
 transformers_5 = version.parse(transformers.__version__) >= version.parse('5.0.0')
 
 
-def _validate_dlrover_flash_checkpoint_api(checkpointer_cls, checkpoint_engine_cls):
-    required_parameters = [
-        (checkpointer_cls.save_checkpoint_to_storage, 'blocking'),
-        (checkpoint_engine_cls.wait_latest_checkpoint, 'max_steps'),
-    ]
-    missing_parameters = []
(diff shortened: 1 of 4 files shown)


tastelikefeet (approved):

swift/trainers/mixin.py
```diff
@@ -64,6 +64,24 @@
 transformers_5 = version.parse(transformers.__version__) >= version.parse('5.0.0')
 
 
+def _validate_dlrover_flash_checkpoint_api(checkpointer_cls, checkpoint_engine_cls):
```
tastelikefeet: _validate_dlrover_flash_checkpoint_api this is a private function( starts with '\_', you can move it to utils.py or just added to sft.py

Issue #9954, opened 2026-08-20
[RFC] Support M2PO for stale-rollout GRPO


primorLee:
### Checklist

- [x] I have searched existing issues and pull requests, and this is a new feature request.

### Feature Request Description

I would like to contribute support for **M2PO (Second-Moment Trust Policy Optimization)** to ms-swift's GRPO training stack.

M2PO was published at [ICLR 2026](https://proceedings.iclr.cc/paper_files/paper/2026/hash/8590b946e4f0c2b1262cb5b047924653-Abstract-Conference.html). It targets off-policy RL with stale rollout data, which is directly relevant to the `fully async rollout` item in the current ms-swift roadmap. Instead of PPO's fixed token-wise clipping interval, the final paper's Algorithm 1 constrains the batch-level second moment of the policy/behavior log-ratio and masks only the largest outliers in the active trust-region quadrants.

Paper and reference implementation:

- Paper: https://arxiv.org/abs/2510.01161
- Reference repository: https://github.com/Infini-AI-Lab/M2PO

#### Proposed scope

1. Add `m2po` as a GRPO `loss_type`, with the paper default `m2_threshold=0.04`.
2. Compute `log(pi_current / pi_behavior)` from `rollout_per_token_logps` when rollout log-probabilities are available.
3. Apply M2 masking only to tokens in the two PPO clipping quadrants: `(A > 0, ratio > 1)` and `(A < 0, ratio < 1)`.
4. Remove the largest squared log-ratio outliers until the second moment of the remaining trust-region tokens is at most the configured threshold.
5. Preserve the paper's denominator: average the masked objective over all valid completion tokens, not only the unmasked tokens.
6. Keep Transformers/HF and Megatron-SWIFT behavior aligned.
7. Add diagnostics such as `m2_before`, `m2_after`, `masked_fraction`, and `trust_region_fraction`.
8. Add deterministic unit tests for masking quadrants, threshold boundaries, padding/empty masks, numerical stability, and loss gradients, plus a small stale-rollout comparison against GRPO.

I already have an independent implementation of the final Algorithm 1 under review in [AgentScope/Trinity-RFT#616](https://github.com/agentscope-ai/Trinity-RFT/pull/616). I would adapt the implementation and its edge-case tests to ms-swift's current `GRPOBatch` and distributed training abstractions rather than copying framework-specific code.

Before implementation, I would appreciate maintainer guidance on three details:

[...]

PR #9965, opened 2026-08-21
feat(grpo): add M2PO for stale-rollout training

primorLee:
# PR type

- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates

# PR information

Implements M2PO (Second-Moment Trust Policy Optimization) for stale-rollout GRPO, following the final ICLR 2026 Algorithm 1. Related RFC: #9954.

## What changed

- adds `loss_type=m2po` and the paper default `m2_threshold=0.04`
- computes `log(pi_current / pi_behavior)` from the actual rollout log-probabilities for vLLM/Megatron paths
- permits `old_per_token_logps` fallback only for synchronous native HF generation; deployment-backed paths fail closed if behavior log-probabilities are missing on any rank
- masks only the active PPO trust-region quadrants and removes the largest squared log-ratio outliers until the remaining batch M2 is within budget
- preserves the paper's original valid-token denominator after masking
- selects once over the complete optimizer batch: HF currently requires `gradient_accumulation_steps=1`; Megatron concatenates all optimizer-step micro-batches before selection
- computes distributed selection across pure data-parallel groups so reconstructed TP/CP replicas are not counted twice
- keeps Transformers/HF and Megatron-SWIFT on the same shared core implementation
- logs `m2_before`, `m2_after`, `masked_fraction`, and `trust_region_fraction`
- rejects unsupported combinations that would silently change the objective: sequence-level IS, extra rollout IS correction without separately retained training-engine behavior log-probabilities, off-policy sequence masking, dual clipping, Liger loss, HF sequence parallel/dynamic loss chunking, and Megatron multi-step reuse
- requires deterministic Megatron mask prepasses; LoRA uses `--lora_dropout 0`, and non-zero model dropout/stochastic depth/BatchNorm/router jitter fail closed
- documents the objective and support boundary in English and Chinese

The reference experiments use `beta=0`, so M2PO defaults to zero KL coefficient when the user does not explicitly set `beta`; an explicit non-zero value remains supported with a warning.

The final paper describes TIS as complementary to M2PO. That composition needs both rollout-engine and training-engine behavior-policy log-probabilities. This implementation currently retains only the rollout behavior tensor, so it rejects an additional `rollout_importance_sampling_mode` instead of double-correcting the same ratio.

## Validation

Validated head `4ad0ed6a464119cc2d8fd3165320e58513ad0cae`:

- Windows targeted M2PO + adjacent rollout-IS regression: **26 passed**
- Linux/Python 3.10 targeted M2PO + adjacent rollout-IS regression: **26 passed**
- repository pre-commit hooks on every changed file: flake8, isort, YAPF, whitespace, EOF, quote, merge-conflict, and mixed-line-ending checks passed
- Python compilation checks for the shared core and both trainer integrations passed
- `git diff --check` passed
- a conflict-free merge tree against current `origin/main` (`174f15e2a812c2669171c1006d8441348ecf35e5`) passed the same 26 tests

The tests cover strict/fallback behavior-policy routing, active trust-region quadrants, final Algorithm 1 equivalence, one optimizer-batch selection across micro-batch splits, original-denominator gradients, masked padding non-finites, invalid inputs, variable distributed token counts, and exclusion of reconstructed CP replicas.

### RTX 4090 CUDA oracle at head

- PyTorch `2.8.0+cu128`, CUDA `12.8`
- valid tokens: 3630; trust-region tokens: 1872; kept trust-region tokens: 938
- M2: `0.307053 -> 0.039919` at threshold `0.04`
- masked fraction: `0.257300`
- CUDA mask matched an independent CPU Algorithm 1 loop and the shared CPU implementation
- analytic-vs-autograd maximum gradient error: `1.16e-10`
- missing deployment behavior log-probabilities and stochastic Megatron prepasses were both rejected as intended

### Real two-node Megatron Context Parallel smoke

The topology/CP fix in parent `60da0d4048e25fa9d916e84ef666085aac737eac` was validated with one NVIDIA A800 80GB (rank 0) and one NVIDIA GeForce RTX 4090 (rank 1):

- aligned runtime: PyTorch `2.8.0+cu128`, CUDA `12.8`, NCCL `2.27.3`, Megatron-Core `0.16.1`, Transformer Engine `2.16.0`
- real two-node CUDA/NCCL world size 2; `all_reduce(rank + 1)` returned `3.0` on both GPUs; both ranks exited 0
- initialized `context_parallel_size=2`; confirmed CP group size 2, DP+CP group size 2, and pure-DP group size 1
- called production `swift.megatron.trainers.utils.reconstruct_tensor_cp` on load-balanced CP shards; both ranks reconstructed `[0.1, 0.2, 0.3, 0.4]`
- negative control with the old DP+CP group produced replica-divergent masks: `[[T,T,F,F], [T,T,T,F]]`
- pure-DP selection produced replica-identical masks: `[[T,T,F,F], [T,T,F,F]]`
- backward gradients matched the analytic shard-local oracle: `[[-0.27629274, 0], [-0.30535069, 0]]`
- metrics matched the oracle: M2 `0.075 -> 0.025`, masked fraction `0.5`, valid/trust/kept counts `4/4/2`

The two-node run used SSH-forwarded loopback sockets plus a process-scoped bind shim because the GPUs are on different private networks. The shim changed only socket port selection; tensor operations, process groups, collectives, reconstruction, masking, loss, and backward used the production PyTorch/NCCL/Megatron/ms-swift paths.

The A800 instance was offline while validating `4ad0ed6`, so the two-node smoke above is intentionally attributed to its exact parent commit rather than presented as a head rerun. Head changes were revalidated through the optimizer-batch distributed tests and the 4090 CUDA oracle above; they do not change CP reconstruction.

## Review status

The behavior-policy decision from #9954 is now resolved fail-closed for deployment-backed rollouts, and the optimizer-batch/parallelism boundaries are explicit. This is ready for maintainer review.


swift/rl_core/data.py
```diff
@@ -324,6 +326,8 @@ class GRPOBatch:
     advantages: Optional[torch.Tensor] = None  # [B, T] per-token (base broadcast minus per-token teacher KL)
     num_items_in_batch: Optional[torch.Tensor] = None  # scalar
     logits_to_keep: Optional[int] = None
+    m2po_mask: Optional[torch.Tensor] = None  # [B, T]
+    m2po_metrics: Optional[Dict[str, torch.Tensor]] = None
```
hjh0119, line 330: The design of GRPOBatch includes only factual data, excluding algorithm-specific data

hjh0119:
Thanks for the contribution! Sorry for the delayed review. I've left a comment

PR #9971, opened 2026-08-23
fix(dataset): honor explicit hub prefixes

YZJF:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Problem

When a caller explicitly selects ModelScope with `ms::dataset-id`, `load_dataset(..., use_hf=True)` still routes the dataset through Hugging Face.

## Root cause

`DatasetSyntax.parse()` represents the explicit ModelScope selection as `False`, but the loader used `dataset_syntax.use_hf or use_hf_default`. Python therefore treated that explicit `False` value as absent and substituted the global default.

## Changes

- Preserve an explicit hub prefix whenever it is present, and use the global default only when no prefix was supplied.
- Add a local regression test that exercises `load_dataset()` with the remote fetch mocked and verifies both prefix parsing and the ModelScope route.

## Experiment results

Baseline RED: the regression test observed `True` for `ms::example-org/private-dataset` when `use_hf=True`; it expected `False`.

GREEN:

- `python -m unittest discover -s tests/general -p 'test_dataset_source.py' -v` — 2 passed
- `python -m py_compile swift/dataset/loader.py tests/general/test_dataset_source.py`
- `pre-commit run --all-files`

## AI assistance

AI assistance was used for investigation and drafting. The production change and the tests above were executed against the current repository checkout.

### DIFF
--- a/swift/dataset/loader.py
+++ b/swift/dataset/loader.py
@@ -325,7 +325,7 @@ def load_dataset(
         use_hf_default = True if use_hf_hub() else False
     for dataset in datasets:
         dataset_syntax = DatasetSyntax.parse(dataset)
-        use_hf = dataset_syntax.use_hf or use_hf_default
+        use_hf = dataset_syntax.use_hf if dataset_syntax.use_hf is not None else use_hf_default
         # compat dataset_name
         if dataset_syntax.dataset in DATASET_MAPPING:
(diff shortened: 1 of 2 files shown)


tastelikefeet (approved):

tastelikefeet:
Hi, can you lint the code by `pre-commit run --all-files` pls

PR #9972, opened 2026-08-24
Preserve Qwen3.5 linear attention FP32 weights

### LINKED ISSUE #9412: Qwen3.5 BF16 training didn't preserve some F32 weights in original model checkpoint
### Checklist / 检查清单

### Bug Description / Bug 描述

The current SFT path in `ms-swift` for Qwen3.5 models didn't keep some explicitly preserved F32 weights in publicly available model checkpoints released by the Qwen Team.
Specifically the `model.language_model.layers.*.linear_attn.out_proj.weight` would be saved as BF16 tensors instead of F32 in the original ckpt.

### How to Reproduce / 如何复现

Run a demostrative SFT job for Qwen3.5 series models:
```shell
swift sft \
    --model Qwen/Qwen3.5-0.8B \
    --tuner_type full \
    --dataset "${DATA_DIR}/train.jsonl" \
    --val_dataset "${DATA_DIR}/val.jsonl" \
    --split_dataset_ratio 0 \
    --torch_dtype bfloat16 \
    --num_train_epochs 1 \
    --per_device_train_batch_size 1 \
    --per_device_eval_batch_size 1 \
    --learning_rate 1e-5 \
    --gradient_accumulation_steps 1 \
    --eval_strategy steps \
    --eval_steps 1 \
    --save_strategy steps \
    --save_steps 1 \
    --save_total_limit 5 \
    --load_best_model_at_end true \
    --metric_for_best_model loss \
    --greater_is_better false \
    --logging_steps 5 \
    --max_length 14336 \
    --truncation_strategy left \
    --output_dir output \
    --warmup_ratio 0.05 \
    --weight_decay 0.05 \
    --dataloader_num_workers 16 \
    --dataset_num_proc 16 \
    --use_chat_template true \
    --model_author Seas0 \
    --model_name simple-classifier \
    --report_to wandb tensorboard \
    --deepspeed zero2
# Manually stop the training after one training step and weight writing...
```
Then check the difference between the original and the saved weights: 
```shell
cat >/tmp/lssafetensors <<'EOF'
#!/bin/sh

set -eu

usage() {
  cat <<'USAGE' >&2
Usage: lssafetensors <path-or-url>

Print the safetensors header JSON metadata.
Supports:
  - local files
  - HTTP/HTTPS URLs via Range requests (useful for web streaming setups)
USAGE
}

if [ "$#" -ne 1 ]; then
  usage
  exit 1
fi

src=$1

read_local() {
  f=$1
  header_len=$(od -An -t u8 -N8 "$f" | tr -d '[:space:]')
  case $header_len in
    ''|*[!0-9]*)
      echo "failed to parse safetensors header length from local file: $f" >&2
      exit 1
      ;;
  esac
  dd if="$f" bs=1 skip=8 count="$header_len" status=none | jq '.'
}

read_remote() {
  url=$1

  prefix=$(curl -fsSL --range 0-7 "$url") || {
    echo "failed to fetch safetensors prefix from URL: $url" >&2
    exit 1
  }

  header_len=$(printf '%s' "$prefix" | od -An -t u8 | tr -d '[:space:]')
  case $header_len in
    ''|*[!0-9]*)
      echo "failed to parse safetensors header length from remote URL: $url" >&2
      exit 1
      ;;
  esac

  end=$((header_len + 7))
  curl -fsSL --range "8-$end" "$url" | jq '.'
}

case $src in
  http://*|https://*) read_remote "$src" ;;
  *)
    if [ ! -f "$src" ]; then
      echo "file not found: $src" >&2
      exit 1
    fi
    read_local "$src"
    ;;
esac
EOF
chmod +x /tmp/lssafetensors

/tmp/lssafetensors ${MODELSCOPE_CACHE}/models/Qwen/Qwen3.5-0.8B/model.safetensors-00001-of-00001.safetensors | jq -S . >/tmp/before-sft.json
/tmp/lssafetensors output/v0-*/checkpoint-1/model.safetensors | jq -S . >/tmp/after-sft.json
diff -u /tmp/{before,after}-sft.json
```
The diff output would show something like this:
```diff
...
   "model.language_model.layers.*.linear_attn.norm.weight": {
     "data_offsets": [
...
     ],
-    "dtype": "F32",
+    "dtype": "BF16",
     "shape": [
       128
     ]
   },
...
```

_No response_

liuhao-labs:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Preserve the Qwen3.5 GatedDeltaNet parameters that the published checkpoint
stores in FP32 when ms-swift requests an FP16 or BF16 model load.

Transformers' Qwen3.5 dense and MoE pretrained classes currently have no
`_keep_in_fp32_modules_strict` policy. A uniform `from_pretrained` dtype thus
casts `linear_attn.A_log` and `linear_attn.norm.weight`, and a later
`save_pretrained` persists the lower precision. The published
Qwen3.5-0.8B checkpoint has 36 FP32 tensors, exactly those two suffixes across
18 layers; `linear_attn.out_proj.weight` is BF16 and remains a control.

This patch merges the two suffixes into the relevant Transformers pretrained
class only while the synchronous dense, MoE, or embedding loader runs. It
preserves existing policy entries, restores the exact prior local or inherited
class state on success or exception, and serializes overlapping Swift policy
contexts with an `RLock`. The loaded model instance retains the policy and the
materialized FP32 tensors after the class state is restored.

Fixes #9412

## Experiment results

Minimal unmodified-main CPU reproduction with the official Qwen3.5-0.8B
checkpoint and requested FP16:

```text
original checkpoint: 36/36 target tensors F32
loaded model:          36/36 target tensors FLOAT16
saved checkpoint:      36/36 target tensors F16
controls:              18/18 out_proj tensors FLOAT16/F16
```

The same oracle after this patch:

```text
loaded model:          36/36 target tensors FLOAT32
saved checkpoint:      36/36 target tensors F32
controls:              18/18 out_proj tensors FLOAT16/F16
```

Targeted CPU/offline regression command:

```bash
CUDA_VISIBLE_DEVICES='' PYTHONDONTWRITEBYTECODE=1 \
HF_HUB_OFFLINE=1 TRANSFORMERS_OFFLINE=1 MODELSCOPE_OFFLINE=1 \
PYTHONPATH=/workspace/ms-swift \
python -m pytest -p no:cacheprovider -q \
  tests/general/test_qwen3_5_fp32_weights.py
```

Result on Transformers 5.15.1: `3 passed, 9 subtests passed`. The same three
tests plus nine subtests pass on 5.2.0 and 5.3.0. All three tests skip on 5.0.0
and 5.1.0 because those wheels do not expose Qwen3.5.

The regression covers tiny dense, MoE, and embedding FP16/BF16 load/save;
exact policy restoration on repeated, nested, exception, and overlapping
two-thread paths; a real `Qwen3_5Loader -> Qwen2VLLoader -> ModelLoader` FP16
forward/backward; and CPU-only optional-kernel isolation. The repository runner
also reports `SUCCESS (Runs=3, success=3, failures=0, errors=0, skipped=0)`.

A CPU-only Swift Trainer save/resume probe trained only the two FP32 targets.
Step 1 and resumed step 2 had finite loss and gradient norm, both targets
changed at each step, and both checkpoints retained two F32 targets plus 36 F16
ordinary tensors. This is correctness evidence, not a full-parameter CPU
training claim.

Formatting and static checks:

```bash
python -m flake8 swift/model/models/qwen.py \
  tests/general/test_qwen3_5_fp32_weights.py
python -m isort --check-only --diff swift/model/models/qwen.py \
  tests/general/test_qwen3_5_fp32_weights.py
python -m yapf --diff swift/model/models/qwen.py \
  tests/general/test_qwen3_5_fp32_weights.py
git diff --check
```

All passed.

GPU SFT, two-GPU ZeRO-2, GPU loss/gradient checks, related GPU model suites, and
the repository complete suite were not run. The lab's loaded NVIDIA module is
535.288.01 while the on-disk module is 580.173.02, so the mandatory GPU gate is
closed. No throughput or GPU-memory benchmark is claimed; retaining 36 small
tensors in FP32 is expected to add memory, but it was not measured.

Compatibility risk: a direct Transformers load running concurrently outside
Swift can observe the temporary class policy. The existing ms-swift model
loader already has the same process-level concurrency boundary while it
temporarily patches `PreTrainedModel.from_pretrained`; overlapping Swift policy
contexts themselves are serialized and tested. A future Transformers release
that declares the policy upstream should supersede this compatibility patch.

Rollback is a direct revert of the Qwen loader changes and deletion of the new
test; there is no checkpoint-format migration.


hjh0119:
can you confirm that 3.6 / 3.8 series models have the same issue?

hjh0119 (approved):

liuhao-labs:
No. I checked all currently registered Qwen3.6/Qwen3.8 checkpoints (dense and MoE, FP8 and non-FP8) on Hugging Face and ModelScope. Their published checkpoints store `linear_attn.A_log` and `linear_attn.norm.weight` in BF16:

- Qwen3.6 dense: 48 + 48 tensors; MoE: 30 + 30
- Qwen3.8 dense: 48 + 48 tensors; MoE: 69 + 69

The FP8 variants use the same BF16 dtype for these target tensors. In contrast, Qwen3.5-0.8B stores both targets in FP32 (18 + 18 tensors).

Because Qwen3.6/Qwen3.8 reuse the Qwen3.5 pretrained classes and ms-swift loaders, the current class-wide strict policy would incorrectly promote their checkpoint-backed BF16 tensors to FP32. I have revised the patch locally so each suffix is enabled only when all expected tensors for that suffix are F32 in the selected local safetensors checkpoint. I also added BF16 no-upcast coverage for dense, MoE, and embedding loaders under requested FP16 and BF16; Transformers 5.2, 5.3, and 5.15.1 all pass with 5 tests and 17 subtests.

The amendment is locally validated and ready to push. The verification used complete published safetensors metadata and tiny CPU loader tests; I did not fully load the large models or run the real FP8/GPU paths.

hjh0119:
can you merge main branch to rerun citest since it's fixed in #9974

liuhao-labs:
Done. I merged the current main branch (40d71be51, including #9974) and pushed the new head db02bd3f2. This also includes the checkpoint-aware Qwen3.5 refinement and the NPU-specific CPU smoke-test skip described above. The updated local version matrix and repository runner are green; GitHub checks should now rerun on the new head.

liuhao-labs:
The three new workflow runs were created for db02bd3f2, but GitHub marked all of them as action_required with no jobs, so they are waiting for upstream approval before they can run: lint #32798657031, citest #32798657037, and citest-npu #32798657035.

PR #10012, opened 2026-08-31
Fix exact token-in/token-out consistency in agentic multi-step rollouts

P90-RushB:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Problem

Agentic multi-step rollouts can decode sampled assistant token IDs into text and later re-encode that text when constructing subsequent inference history or training inputs. Tokenization is not generally bijective, so this round trip can change the token sequence.

As a result, the sequence used for training may differ from the sequence actually sampled by the inference engine. This can happen across scheduler mutations, tool observations, deterministic response prefixes, agentic continuations, and multimodal chunk reconstruction.

This mismatch is especially harmful for agentic reinforcement learning because the policy may be updated using a trajectory that no longer exactly matches the rollout generated by the policy.

ms-swift already propagates `response_token_ids` and `response_loss_mask` through several rollout and training paths, including handling for response prefixes. This PR preserves those existing interfaces and closes the remaining end-to-end consistency gaps in agentic multi-step rollouts.

## Why exact token-in/token-out matters

The verl agent-loop documentation discusses the same distinction between chat-completion-based rollouts and exact token-in/token-out generation:

https://verl.readthedocs.io/en/latest/advance/agent_loop.html#chat-completion-vs-token-in-token-out

It notes that tool parsing and decode-encode round trips can cause the token IDs reconstructed from the final message history to differ from the concatenated prompt and response IDs produced during rollout.

For serving, this difference may be tolerable because the rendered text can remain semantically equivalent. For reinforcement learning, however, the token trajectory is part of the training sample. Reconstructing a different sequence can move the training trajectory away from the policy distribution that generated it and can materially harm convergence.

This PR follows the same invariant: engine-sampled token IDs remain the source of truth throughout rollout collection and training-input construction. Text materialization is used only temporarily when scheduler hooks require text.

## Fix

- Keep engine-sampled token IDs as the source of truth across colocate and server multi-turn rollouts.
- Preserve token-backed assistant history across scheduler hooks and agentic continuations.
- Represent deterministic response prefixes and scheduler-added content with aligned loss masks.
- Keep scheduler-provided token IDs, loss masks, and rollout log probabilities aligned.
- Reuse exact response tokens when rebuilding multimodal training chunks.
- Forward the template and tokenizer through Ray Megatron GRPO/GKD scheduler setup.
- Avoid duplicating template separators already present in token-backed responses.
- Add validation for invalid response loss masks.
- Add regression coverage for non-bijective tokenization, prefixes, continuations, tool calls, server/colocate paths, multimodal chunks, and Ray schedulers.

## AgentArk Snake experiment

I found and investigated this issue while training the Snake task with AgentArk, an open-source environment for agentic and multimodal reinforcement learning:

- AgentArk repository: https://github.com/P90-RushB/AgentArk
- AgentArk ms-swift Snake training tutorial: https://github.com/P90-RushB/AgentArk/blob/main/integrations/ms_swift/tutorial/README.zh-CN.md

I ran two comparison experiments using the AgentArk Snake training pipeline:

1. The upstream ms-swift implementation without the complete exact-token fix.
2. The patched ms-swift implementation from this PR.

### Upstream ms-swift without the fix

In the upstream run, `train/reward` progressively decays. Toward the end of the plotted training window, most reward curves have collapsed close to zero.

AgentArk Snake training with upstream ms-swift

### ms-swift with the exact-token fix

With the inference-generated token IDs preserved throughout the multi-step rollout and training-input reconstruction, reward remains positive. Multiple curves continue to improve instead of collapsing toward zero.

AgentArk Snake training with exact token-in/token-out preservation

The two screenshots use different y-axis ranges, so the comparison should be based on the labeled reward values and overall trends rather than only the visual height of the curves.

In this AgentArk Snake experiment, the upstream run shows severe reward degradation, while the patched run maintains a usable training signal and shows substantially healthier reward behavior.

This experiment is not intended as a general benchmark of AgentArk or ms-swift. It provides practical evidence that losing exact token-in/token-out consistency can materially degrade agentic RL training.

## Test coverage

The patch adds focused regression tests for:

- non-bijective decode/re-encode behavior;
- exact sampled response-token preservation;
- deterministic and scheduler-added response prefixes;
- aligned response loss masks;
- invalid response loss-mask values and lengths;
- agentic continuations;
- tool-call observations and deterministic tool-result tokens;
- colocate and server rollout paths;
- multimodal training-chunk reconstruction;
- template separator deduplication;
- Ray Megatron GRPO and GKD scheduler initialization.

The tests are implemented as `unittest.TestCase` classes so they can be collected by the existing ms-swift test runner:

- `tests/rollout/test_exact_token_io.py`: 12 test methods
- `tests/rollout/test_server_exact_token_io.py`: 4 test methods

## Local validation

- Python 3.12 static compilation passes for all changed source and test files.
- Git diff whitespace validation passes.
- Full runtime execution of the tests was not available in the local preparation environment because it did not contain a compatible complete torch/transformers ms-swift test stack. The PR is opened as a draft for upstream CI validation.

### DIFF
--- a/swift/ray/megatron/grpo_trainer.py
+++ b/swift/ray/megatron/grpo_trainer.py
@@ -105,7 +105,11 @@ class GRPOTrainer(BaseRayTrainer):
             if scheduler_cfg not in multi_turns:
                 raise ValueError(f'Unknown multi_turn_scheduler: {scheduler_cfg!r}; '
                                  f'available: {list(multi_turns)}')
-            scheduler_kwargs = {'max_turns': self._max_turns}
+            scheduler_kwargs = {
+                'max_turns': self._max_turns,
+                'tokenizer': self.template.tokenizer,
+                'template': self.template,
+            }
             gym_env = getattr(args, 'gym_env', None)
             if gym_env is not None:
                 scheduler_kwargs['gym_env'] = gym_env
--- a/swift/template/base.py
+++ b/swift/template/base.py
@@ -25,7 +25,8 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, 
 
 from swift.utils import Processor, ProcessorMixin, get_env_args, get_logger, remove_response, retry_decorator, to_device
 from .template_inputs import StdTemplateInputs, TemplateInputs
-from .utils import Context, ContextType, StopWordsCriteria, fetch_one, findall, get_last_user_round, split_str_parts_by
+from .utils import (Context, ContextType, StopWordsCriteria, fetch_one, findall, get_last_user_round,
+                    get_token_backed_response_ids, split_str_parts_by)
 from .vision_utils import _check_path, load_audio, load_batch, load_image, rescale_image
 
 logger = get_logger()
@@ -964,22 +965,10 @@ class Template(ProcessorMixin):
     def _tokenize(self, context, **kwargs):
         return self.tokenizer(context, return_attention_mask=False, add_special_tokens=False, **kwargs)['input_ids']
 
-    @staticmethod
-    def _get_token_backed_response_ids(response: Context) -> Optional[List[int]]:
-        if isinstance(response, dict):
-            token_ids = response.get('token_ids', response.get('input_ids'))
-        elif isinstance(response, list) and (not response or isinstance(response[0], int)):
-            token_ids = response
-        else:
-            return None
-        if not isinstance(token_ids, list) or any(not isinstance(token_id, int) for token_id in token_ids):
-            return None
-        return list(token_ids)
-
     def _remove_response_separator_overlap(self, response: Context,
                                            extra_context_list: Optional[List[Context]]) -> List[Context]:
         """Keep sampled terminal tokens while removing their overlap from a template separator."""
-        response_ids = self._get_token_backed_response_ids(response)
+        response_ids = get_token_backed_response_ids(response)
         if not response_ids or not extra_context_list:
             return extra_context_list or []
         if any(not isinstance(context, str) for context in extra_context_list):
--- a/swift/template/base.py
+++ b/swift/template/base.py
@@ -964,6 +964,37 @@ class Template(ProcessorMixin):
     def _tokenize(self, context, **kwargs):
         return self.tokenizer(context, return_attention_mask=False, add_special_tokens=False, **kwargs)['input_ids']
 
+    @staticmethod
+    def _get_token_backed_response_ids(response: Context) -> Optional[List[int]]:
+        if isinstance(response, dict):
+            token_ids = response.get('token_ids', response.get('input_ids'))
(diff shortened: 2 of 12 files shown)


swift/ray/megatron/grpo_trainer.py
```diff
@@ -105,7 +105,11 @@ def _prepare_multi_turn(self) -> None:
             if scheduler_cfg not in multi_turns:
                 raise ValueError(f'Unknown multi_turn_scheduler: {scheduler_cfg!r}; '
                                  f'available: {list(multi_turns)}')
-            scheduler_kwargs = {'max_turns': self._max_turns}
+            scheduler_kwargs = {
+                'max_turns': self._max_turns,
+                'tokenizer': self.template.tokenizer,
+                'template': self.template,
+            }
```
hjh0119, line 112: Could you also pass template in `_prepare_scheduler` for both `RolloutTrainerMixin` (swift/rlhf_trainers/rollout_mixin.py) and `MegatronRolloutMixin` (swift/megatron/trainers/rollout_mixin.py)? In colocate mode infer_engine is None, so without it the response_prefix injection in get_response_token_data silently doesn't happen on these paths.

swift/template/base.py
```diff
@@ -964,6 +964,37 @@ def _split_special_tokens(context_list: List[Context],
...
+    def _get_token_backed_response_ids(response: Context) -> Optional[List[int]]:
+        if isinstance(response, dict):
+            token_ids = response.get('token_ids', response.get('input_ids'))
+        elif isinstance(response, list) and (not response or isinstance(response[0], int)):
+            token_ids = response
+        else:
+            return None
+        if not isinstance(token_ids, list) or any(not isinstance(token_id, int) for token_id in token_ids):
+            return None
+        return list(token_ids)
```
hjh0119: There are actually 3 hand-rolled copies of this token-backed detection (template/base.py + two inline copies in rollout/multi_turn.py). Moving it to a shared util would also let the rollout side reuse it instead of duplicating the checks.

hjh0119 (approved):

swift/ray/megatron/grpo_trainer.py
P90-RushB: Addressed in 55c3a98: both RolloutTrainerMixin and MegatronRolloutMixin now pass the template alongside the tokenizer when constructing a scheduler. I also added regression coverage for the RolloutTrainerMixin colocate path.

swift/template/base.py
P90-RushB: Addressed in 55c3a98: the token-backed response detection now lives in get_token_backed_response_ids in swift/template/utils.py and is reused by Template and both rollout call sites. Direct utility coverage was added as well.

hjh0119:
Thanks for catching this!

PR #10029, opened 2026-09-02
Support mixed text and raw-token rollout content in Swift template

kerbeans:
## Summary

Support mixed rollout response content in `Template._swift_encode()`.

Rollout responses may contain multiple content components, including a text segment followed by a raw-token segment, for example:

```python
[
    "<think>...</think><tool_call>...</tool_call>",
    {
        "loss_scale": [...],
        "token_ids": [...]
    }
]
```

The current Swift template encoding path assumes that a list whose final item is not a string can be treated directly as `List[int]`. For mixed content such as `List[str, dict]`, this passes the entire list to `tokenizer.decode()`, causing a type error.

This change recognizes a trailing raw-token component and uses its `token_ids` only for suffix inspection. The original response representation is preserved for template encoding, so raw-token rollout semantics are unchanged.

## Tests

- `tests/general/test_template_mixed_rollout.py`: 5 passed
- Existing exact-token rollout tests: 33 passed; one unrelated existing test requires a missing `trainer_name` fixture

### DIFF
--- a/docs/source/Instruction/Command-line-parameters.md
+++ b/docs/source/Instruction/Command-line-parameters.md
@@ -157,6 +157,7 @@
   - 注意：eos_token会在输出respsone中被删除，额外停止词会在输出中保留。
 - logprobs: 是否输出logprobs，默认为False。
 - top_logprobs: 输出top_logprobs的数量，默认为None。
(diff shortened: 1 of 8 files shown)


hjh0119:
When would this format happen? Could you create an issue with more details?

kerbeans:
> When would this format happen? Could you create an issue with more details?
I have created issue #10077.
This was happened when student rollout tool calls in GKD training.

PR #10036, opened 2026-09-04
fix: default `dataloader_persistent_workers=True` to avoid per-eval worker cold start

### LINKED ISSUE #10016: Data loaders somehow pull in CUDA
### Checklist / 检查清单

### Bug Description / Bug 描述

In current main (5e43bdff0af1dee55b014e0e9b9eb9ba25e229f3), on Python 3.14, it seems the data loaders somehow pull in CUDA. I am running swift without `--dataloader_multiprocessing_context`, and I believe the default of `forkserver` gets used.

I am running, essentially,

```
swiftwrap/.venv/bin/swift sft \
    --model Qwen/Qwen3.5-0.8B-Base \
    --tuner_type lora \
    --dataset train.jsonl \
    --val_dataset val.jsonl \
    --dataloader_num_workers 6 \
    --eval_on_start true \
    --eval_strategy steps \
    --eval_steps 15 \
    --save_strategy steps \
    --save_steps 15
```

A few observations:

1. I can see long (20-25 second) pauses after a run of training steps has finished and before evaluation begins. During this I can see "[INFO:swift] Successfully registered `.../dataset_info.json`" printed 6 times, with 1-2 second between each print. There does not seem to be a pause when moving from evaluation to training.
2. In `nvtop` (or nvidia-smi), I can observe that the data loader workers show up as Compute processes, seemingly indicating they somehow decided to instantiate a CUDA context.
3. If I add `--dataloader_persistent_workers true ` to the command line, the pauses disappear. However, the workers still show as CUDA processes in nvidia-smi.

I hypothesize that something the workers receive via pickle cause them to pull in CUDA and probably other slow-to-import libraries.

I assume them pulling in CUDA is some kind of a bug. As for each of them importing torch (without CUDA), which I assume is intended behavior, I think adding `torch` to forkserver preloaded libraries (`multiprocessing.set_forkserver_preload`) might be a good idea?

I actually would also suspect, without inspecting, that the data loader starting is essentially serialized—a process is spawned, things are pickled to it, perhaps main process waits for it to have done all the expensive imports it needs and to receive the pickle, and then it proceeds to start the next data loader? (This hypothesis may well be wrong; it's not much more than a guess.)

### How to Reproduce / 如何复现

See above. Mostly, observing timing and processes with nvtop.

Here's an idea: Maybe the data loader could do some Python introspection once it's running to verify that it doesn't have CUDA loaded, perhaps by inspecting `sys.modules` or something? That won't fix issues, but it could make them more visible.

zhiwuyazhe-z:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Fixes #10016 (the eval-startup stalls).

On Python 3.14 the default multiprocessing start method is forkserver, and torch's
DataLoader rebuilds its worker pool whenever the eval dataloader is re-created.
Every evaluation therefore pays a cold start: each worker re-imports the full
torch/transformers/swift stack before the first batch. Measured on a Qwen3.5-0.8B
LoRA SFT with `dataloader_num_workers=6` and an 8-sample val set, each eval took
~30 s, almost all of it worker startup.

`dataloader_persistent_workers` already fixes this (the reporter of #10016 also
observed it), but defaults to False. This PR flips the default to True in both
`swift/trainers/arguments.py` and `swift/megatron/arguments/megatron_args.py`.

Trade-off: idle workers keep their imported modules in host memory between
evaluations (on the order of a few hundred MB per worker). Users who prefer the
old behavior can pass `--dataloader_persistent_workers false`.

Note: the other observation in #10016 (worker processes holding CUDA contexts)
was not reproducible with torch 2.14, which resets CUDA state in forked children
(workers report `torch.cuda.is_initialized() == False`); the megatron-side default
flip is the same one-line semantic change but was not run on a megatron cluster.

## Experiment results

Qwen3.5-0.8B-Base LoRA SFT, 40 train / 8 val samples, 6 workers, eval every
3 steps with `eval_on_start`, Python 3.14, torch 2.14, RTX A6000:

| config | eval@0 (cold) | eval@3 | eval@6 |
|--------|---------------|--------|--------|
| main (default False) | 30.2 s | 30.1 s | 29.7 s |
| this PR (default True) | 32.3 s | 0.70 s | 0.70 s |

In the 5-eval version of this run, wall time spent in eval worker startup
dropped from ~45% to ~5%.

### DIFF
--- a/swift/megatron/arguments/megatron_args.py
+++ b/swift/megatron/arguments/megatron_args.py
@@ -515,7 +515,7 @@ class MegatronArguments(RLHFMegatronArgumentsMixin, MegatronTunerMixin):
     train_dataloader_shuffle: bool = True
     dataloader_num_workers: int = 4
     dataloader_pin_memory: bool = True
-    dataloader_persistent_workers: bool = False
+    dataloader_persistent_workers: bool = True
     dataloader_prefetch_factor: int = 2
     dataloader_multiprocessing_context: Optional[Literal['fork', 'spawn', 'forkserver']] = None
     data_sharding: bool = False
--- a/swift/trainers/arguments.py
+++ b/swift/trainers/arguments.py
@@ -49,7 +49,7 @@ class TrainArgumentsMixin:
             and specify the API KEY corresponding to your account through `WANDB_API_KEY`.
         dataloader_num_workers (Optional[int]): The number of subprocesses to use for data loading. Defaults to None.
         dataloader_persistent_workers (bool): If True, the data loader workers will not be shut down after a dataset
-            has been consumed once. Defaults to False.
+            has been consumed once. Defaults to True.
         dataloader_prefetch_factor (Optional[int]): The number of batches loaded in advance by each worker. Defaults
             to None.
         use_liger_kernel (bool): Whether to use the Liger kernel for optimization. Defaults to False.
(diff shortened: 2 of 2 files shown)


tastelikefeet (approved):

dv0:
开启dataloader_persistent_workers ，也能一定程度缓解offload内存暴涨的问题  #9687

PR #10059, opened 2026-09-07
fix(embedding): avoid duplicate WeMM backbone forward

taking-lying-flat:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Calling a WeMM embedding model currently runs its inherited generation forward before the embedding post-forward hook runs the backbone again. The first backbone result and vocabulary logits are discarded.

Replace the post-forward hook with a forward wrapper that calls `self.embedding()` directly and preserves the existing `last_hidden_state` output format. Preserve the forward signature and use `patch_module_forward` to retain Accelerate device dispatch hooks.

## Experiment results

CPU checks used the official WeMM class, the actual loader adaptations, and a tiny Qwen3.5 configuration with Transformers 5.14.1:

| Calls per model invocation | Before | After |
| --- | ---: | ---: |
| Backbone | 2 | 1 |
| LM head | 1 | 0 |

Eight cases passed across eval/train mode, attention dropout 0/0.2, positional/keyword inputs, and an Accelerate device hook. Embeddings, tested parameter gradients, and RNG state exactly match direct `embedding()` calls. Forward signature and device hook execution are preserved.

`uvx pre-commit run --files swift/model/models/qwen.py` and `git diff --check` passed. Full-size GPU and multimodal benchmarks were not run.

### DIFF
--- a/swift/model/models/qwen.py
+++ b/swift/model/models/qwen.py
@@ -6,6 +6,7 @@ import os
 import torch
 import torch.nn.functional as F
 from contextlib import contextmanager
+from functools import wraps
 from importlib import import_module
 from packaging import version
 from PIL import Image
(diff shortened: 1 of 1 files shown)


z0o0ey (approved):
o

PR #10063, opened 2026-09-08
fix(model): rewrite seq_cls architectures to *ForSequenceClassification

### LINKED ISSUE #9704: Qwen3-VL 多标签序列分类（seq_cls, num_labels>2, multi_label_classification）训练后无法用 vLLM 部署
### Checklist / 检查清单

### Bug Description / Bug 描述

ms-swift 版本：3.12.0.dev0 
基座模型：Qwen/Qwen3-VL-8B-Instruct
任务：多模态序列分类（seq_cls）
训练参数：--task_type seq_cls --problem_type multi_label_classification --num_labels 20
微调方式：full sft（也复现过 LoRA）
vLLM 版本：0.11.0

## 问题描述
训练参考脚本：https://github.com/modelscope/ms-swift/blob/release/3.12/examples/train/seq_cls/multi_label/vlm.sh
我用 ms-swift参考上面的脚本 对 **Qwen3-VL-8B** 做了**多标签序列分类**微调（`num_labels=20`，`problem_type=multi_label_classification`），训练本身正常，得到的 checkpoint 里 `score.weight` 形状为 `[20, 4096]`，`config.json` 里也带了 20 项的 `id2label` / `label2id` 和 `problem_type=multi_label_classification`。

**用 `PtEngine` / `swift infer --infer_backend pt` 推理完全正常，能正确输出 20 维概率。**

但问题是：**ms-swift 导出的 checkpoint `config.json` 里 `architectures` 仍然是 `Qwen3VLForConditionalGeneration`（生成架构），而不是分类架构**。这导致下游用 vLLM 部署时无论怎么配都失败：

## 复现步骤

1. 训练：
```bash
swift sft \
  --model Qwen/Qwen3-VL-8B-Instruct \
  --task_type seq_cls \
  --problem_type multi_label_classification \
  --num_labels 20 \
  --dataset <多标签数据集> \
  ...
```

2. 查看导出的 `config.json`：
```json
{
  "architectures": ["Qwen3VLForConditionalGeneration"],   // ← 仍是生成架构
  "model_type": "qwen3_vl",
  "problem_type": "multi_label_classification",
  "id2label": { "0": "LABEL_0", ..., "19": "LABEL_19" },   // 20 项
  "label2id": { "LABEL_0": 0, ..., "LABEL_19": 19 }
}
```

3. 尝试用 vLLM 部署（三种方式全部失败）：

**方式一：覆盖架构为 `Qwen3VLForSequenceClassification`**
```bash
vllm serve ./checkpoint-400/ --runner pooling \
  --hf-overrides '{"architectures": ["Qwen3VLForSequenceClassification"], "problem_type": "multi_label_classification"}'
```
报错（vLLM 该类硬编码为 reranker 的 2 分类头，忽略 num_labels）：
```text
AssertionError: Tried to load weights of size torch.Size([20, 4096])
                to a parameter of size torch.Size([2, 4096])
```

**方式二：不覆盖架构，仅 `--runner pooling`**
```bash
vllm serve ./checkpoint-400/ --runner pooling \
  --hf-overrides '{"problem_type": "multi_label_classification"}'
```
报错（退化成 embedding 模型，没有分类头）：
```text
ValueError: There is no module or parameter named 'score' in Qwen3VLForEmbedding.
```

**方式三：保留 config 里的 `Qwen3VLForConditionalGeneration`**
被当作生成模型加载，`/classify` 端点不可用。

## 期望行为

希望 ms-swift 在导出 seq_cls 多标签模型时能做到以下之一，方便下游 vLLM 部署：

1. **导出时把 `config.json` 的 `architectures` 写成对应的分类架构名**（如 `Qwen3VLForSequenceClassification`），而不是保留基座的 `Qwen3VLForConditionalGeneration`；
2. 或在文档中**明确说明**：Qwen3-VL 这类多模态 seq_cls 模型（num_labels>2）目前**只能用 `pt` 后端部署，不支持裸 vLLM**，并给出 `swift deploy --infer_backend pt` 的推荐用法；
3. 或提供一个**导出为 vLLM 兼容格式**的选项（如 `swift export` 时正确写入分类头与架构名）。

## 疑问

- ms-swift 训练 seq_cls 时，导出 config 的 `architectures` 保留生成架构名是有意为之，还是可以改进为分类架构名？
- `swift deploy --infer_backend vllm` 对多模态多标签 seq_cls 模型是否已支持？如果支持，正确的参数是什么？

## 目前的可用方案

`pt` 后端一切正常：
```python
from swift.llm import PtEngine, BaseArguments
model_path = './checkpoint-400/'
args = BaseArguments.from_pretrained(model_path)
engine = PtEngine(model_path, task_type='seq_cls',
                  num_labels=args.num_labels, problem_type=args.problem_type)
```

感谢！

### How to Reproduce / 如何复现

vllm serve ./checkpoint-400/ --runner pooling \
  --hf-overrides '{"architectures": ["Qwen3VLForSequenceClassification"], "problem_type": "multi_label_classification"}'

_No response_

RerankerGuo:
## Summary

Fixes #9704.

When fine-tuning a VLM or LM with `task_type=seq_cls` (or `reranker`), the seq_cls patcher in `_patch_sequence_classification` monkey-patches a `score` head onto the **generation** model class without swapping the class itself. `transformers.PreTrainedModel.save_pretrained` then writes `model.__class__.__name__` into `config.json['architectures']`, so the on-disk checkpoint advertises e.g. `Qwen3VLForConditionalGeneration` while shipping a `score` head, `num_labels`, `id2label`, and `problem_type=multi_label_classification`.

Downstream vLLM deployment reads `architectures` to pick the model class, finds the generation class, and rejects the checkpoint — even though `PtEngine` inference works correctly. The reporter in #9704 confirmed all four of `score.weight: [20, 4096]`, `id2label`, `label2id`, and `problem_type` are written correctly; only `architectures` is wrong.

## Fix

Rewrite `model.config.architectures` in-place to the matching `*ForSequenceClassification` class at the same place the `score` head is attached, so every save path (trainer `_save_model`, `save_checkpoint`, export-quant, peft merge-and-unload) writes the right value automatically. The existing `vllm_engine.py:347-353` `arch_mapping` override is left in place — it remains useful for users who already have bad checkpoints on disk.

The new helper `_seq_cls_architectures` handles the suffix rewrite idempotently and leaves unknown / custom architectures untouched (so we never advertise a class that does not exist in transformers).

## Mapping coverage

| Source class | Rewritten to |
| --- | --- |
| `*ForConditionalGeneration` (Qwen3-VL, Qwen2-VL, InternVL, LLaVA, GLM4V, …) | `*ForSequenceClassification` |
| `*ForCausalLM` (Qwen2, Llama, Mistral, Mixtral, Yi, …) | `*ForSequenceClassification` |
| Already `*ForSequenceClassification` | unchanged (idempotent) |
| Custom / unknown suffix | unchanged |
| `[]` / `None` | unchanged |

## Tests

Added `TestSeqClsArchitecturesRewrite` in `tests/general/test_model.py` covering:
- generation → seq_cls rewrite for both `ForCausalLM` and `ForConditionalGeneration` suffixes
- idempotence on already-seq_cls classes
- multi-arch lists
- empty / `None` inputs
- unknown / custom architectures left untouched

Local sanity checks (10/10 PASS) and syntax checks on both files PASS. Full ms-swift unit / smoke CI is needed to validate end-to-end on real seq_cls training runs; that will run on the upstream CI after this PR is opened.

## Out of scope

- Megatron seq_cls `architectures` rewrite — Megatron has its own seq_cls head code path (`swift/megatron/utils/convert_utils.py:283`, `swift/megatron/model/utils.py:47`); happy to follow up with a separate PR if maintainers want it.
- `vllm_engine.py:347-353` `arch_mapping` cleanup — kept for backward-compat with existing bad checkpoints.

### DIFF
--- a/swift/model/patcher.py
+++ b/swift/model/patcher.py
@@ -249,6 +249,29 @@ def transformers_seq_cls_forward(self, *args, origin_forward, padding_side=None,
     )
 
 
+def _seq_cls_architectures(arch_list):
+    """Rewrite architectures to the matching ``*ForSequenceClassification`` class.
+
+    Used by :func:`_patch_sequence_classification` so that the on-disk
+    ``config.json`` produced by ``PreTrainedModel.save_pretrained`` references
+    the seq_cls architecture (e.g. ``Qwen3VLForSequenceClassification``) instead
+    of the generation architecture that the model class is actually an instance
+    of. Without this, downstream vLLM deployment fails because the checkpoint
+    advertises ``Qwen3VLForConditionalGeneration`` while shipping a
+    ``score`` head and ``num_labels`` / ``problem_type`` fields — see #9704.
+    """
+    if not arch_list:
+        return arch_list
+    res = []
+    for arch in arch_list:
+        if arch.endswith('ForConditionalGeneration'):
+            arch = arch[:-len('ForConditionalGeneration')] + 'ForSequenceClassification'
+        elif arch.endswith('ForCausalLM'):
+            arch = arch[:-len('ForCausalLM')] + 'ForSequenceClassification'
+        res.append(arch)
+    return res
+
+
 def _patch_sequence_classification(model, model_meta):
     hidden_size = HfConfigFactory.get_config_attr(model.config, 'hidden_size')
     initializer_range = HfConfigFactory.get_config_attr(model.config, 'initializer_range')
@@ -273,6 +296,12 @@ def _patch_sequence_classification(model, model_meta):
         return transformers_seq_cls_forward(self, *args, origin_forward=origin_forward, **kwargs)
 
     lm_head_model.forward = MethodType(new_forward, lm_head_model)
+
+    # Align the on-disk `architectures` with the task. PreTrainedModel.save_pretrained
+    # writes `model.__class__.__name__` for `architectures`, but the seq_cls patcher
+    # monkey-patches a `score` head onto the generation class without swapping it,
+    # so the saved checkpoint would otherwise advertise the wrong class (see #9704).
+    model.config.architectures = _seq_cls_architectures(getattr(model.config, 'architectures', None))
 
 
 @contextmanager
--- a/tests/general/test_model.py
+++ b/tests/general/test_model.py
@@ -48,6 +48,53 @@ class TestMolmo2Registration(unittest.TestCase):
         self.assertIn(TemplateType.molmo2, TEMPLATE_MAPPING)
 
 
+class TestSeqClsArchitecturesRewrite(unittest.TestCase):
+    """Regression coverage for #9704.
+
+    The seq_cls / reranker patcher monkey-patches a ``score`` head onto the
+    generation model class without swapping the class itself, so
+    ``PreTrainedModel.save_pretrained`` would otherwise write the wrong
+    ``architectures`` to ``config.json`` and break downstream vLLM deployment.
+    """
+
+    def test_seq_cls_architectures_rewrite(self):
+        from swift.model.patcher import _seq_cls_architectures
+
+        # Generation -> SequenceClassification
+        self.assertEqual(_seq_cls_architectures(['Qwen2ForCausalLM']), ['Qwen2ForSequenceClassification'])
+        self.assertEqual(
+            _seq_cls_architectures(['Qwen3VLForConditionalGeneration']),
+            ['Qwen3VLForSequenceClassification'])
+        self.assertEqual(_seq_cls_architectures(['LlamaForCausalLM']), ['LlamaForSequenceClassification'])
+        self.assertEqual(
+            _seq_cls_architectures(['Qwen2VLForConditionalGeneration']),
+            ['Qwen2VLForSequenceClassification'])
+
+        # Already-seq_cls class is preserved (idempotent).
+        self.assertEqual(
+            _seq_cls_architectures(['BertForSequenceClassification']),
+            ['BertForSequenceClassification'])
+
+        # Multi-arch list: only the matching suffix is rewritten.
+        self.assertEqual(
+            _seq_cls_architectures(['FooForCausalLM', 'BarForConditionalGeneration']),
+            ['FooForSequenceClassification', 'BarForSequenceClassification'])
+
+        # Empty / None inputs are returned unchanged.
+        self.assertEqual(_seq_cls_architectures([]), [])
+        self.assertIsNone(_seq_cls_architectures(None))
+
+    def test_seq_cls_architectures_unknown_suffix(self):
+        """Unknown suffixes (custom architectures) are left alone — we don't
+        silently invent a class name that may not exist in transformers."""
+        from swift.model.patcher import _seq_cls_architectures
+
+        self.assertEqual(_seq_cls_architectures(['MyCustomLMHeadModel']), ['MyCustomLMHeadModel'])
+        self.assertEqual(
+            _seq_cls_architectures(['MyCustomLMHeadModel', 'LlamaForCausalLM']),
+            ['MyCustomLMHeadModel', 'LlamaForSequenceClassification'])
+
+
 if __name__ == '__main__':
     test_qwen2()
     # test_modelscope_hub()


Excelius-Wang (changes requested):
This still saves the generation architecture to `config.json`. `PreTrainedModel.save_pretrained()` overwrites `config.architectures` with the model's class name, so the assignment in `_patch_sequence_classification` does not survive saving.

I reproduced this on CPU with Transformers 5.12.1 and a randomly initialized tiny Qwen2 model:

```python
config = Qwen2Config(
    vocab_size=32, hidden_size=16, intermediate_size=32,
    num_hidden_layers=1, num_attention_heads=2, num_key_value_heads=2,
    num_labels=2, architectures=["Qwen2ForCausalLM"],
)
model = Qwen2ForCausalLM(config)
_patch_sequence_classification(model, SimpleNamespace(model_arch=None))
# model.config.architectures is now ["Qwen2ForSequenceClassification"]
model.save_pretrained(output_dir)
# config.json contains ["Qwen2ForCausalLM"] again
```

Could the fix cover the save path and add a regression that reads the saved `config.json`? The two new helper tests pass, but this save test fails, leaving the checkpoint metadata problem described in the PR unresolved. No model downloads or training were needed to reproduce it.

tastelikefeet (approved):

tastelikefeet:
Please lint by `pre-commit run --all-files`

swift/model/patcher.py
```diff
@@ -274,6 +297,12 @@ def new_forward(self, *args, **kwargs):
 
     lm_head_model.forward = MethodType(new_forward, lm_head_model)
 
+    # Align the on-disk `architectures` with the task. PreTrainedModel.save_pretrained
+    # writes `model.__class__.__name__` for `architectures`, but the seq_cls patcher
+    # monkey-patches a `score` head onto the generation class without swapping it,
+    # so the saved checkpoint would otherwise advertise the wrong class (see #9704).
+    model.config.architectures = _seq_cls_architectures(getattr(model.config, 'architectures', None))
```
hjh0119, line 304: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L3356 still overwrites; add tests for actual save paths

RerankerGuo:
Thanks for the clear reproduction — you're right, `PreTrainedModel.save_pretrained()` resets `config.architectures` to `model.__class__.__name__` right before serializing, so the in-memory assignment never reached `config.json`.

Fixed in baaf078 by restoring the seq_cls names inside `PretrainedConfig.save_pretrained`, the last hook before the config is written:

- `_patch_save_pretrained_architectures` (swift/model/patcher.py:275) wraps the patched model's `save_pretrained` and re-applies the seq_cls `architectures` only when transformers has just reset them to the class name, so a value assigned after patching is preserved.
- New regression test `test_seq_cls_architectures_saved_config` (tests/general/test_model.py:93) patches a tiny Qwen2 model, saves it, and asserts on `architectures` read back from the written `config.json`, plus a round-trip through `AutoConfig.from_pretrained`.

Verified locally on CPU with transformers 5.12.1 using your repro:
- before: `config.json` → `["Qwen2ForCausalLM"]`
- after: `config.json` → `["Qwen2ForSequenceClassification"]`
- unknown suffixes (e.g. `MyCustomLMHeadModel`) are still left untouched, no extra keys leak into `config.json`, and an unpatched model is unaffected.

Also fixed the yapf formatting that was failing the `lint` job. `lint` and the first `unittest` job are green; the second `unittest` is still queued on the runner.

swift/model/patcher.py
RerankerGuo: Agreed, that is exactly the overwrite I missed — `PreTrainedModel.save_pretrained` resets `config.architectures` to `model.__class__.__name__` (modeling_utils.py:109 in v5.12.1) before the config is serialized, so the in-memory assignment never reached `config.json`.

Fixed in baaf078:

- `_patch_save_pretrained_architectures` (swift/model/patcher.py:275) restores the seq_cls names inside `PretrainedConfig.save_pretrained`, i.e. after transformers overwrites them and before the file is written.
- `test_seq_cls_architectures_saved_config` (tests/general/test_model.py:93) now covers the actual save path: it patches a tiny Qwen2 model, calls `save_pretrained`, and asserts on the `architectures` read back from the written `config.json`, plus a round-trip through `AutoConfig.from_pretrained`.

Verified locally on CPU with transformers 5.12.1: before the fix `config.json` contained `["Qwen2ForCausalLM"]`, now `["Qwen2ForSequenceClassification"]`; unknown suffixes are left untouched and no extra keys leak into the config. `lint` and both `unittest` jobs are green on the latest commit.

swift/model/patcher.py
```diff
@@ -249,6 +249,70 @@ def transformers_seq_cls_forward(self, *args, origin_forward, padding_side=None,
...
+            return config_save_pretrained(*args, **kwargs)
+
+        config.save_pretrained = _config_save_pretrained
+        try:
+            return model_save_pretrained(save_directory, *args, **kwargs)
+        finally:
+            config.__dict__.pop('save_pretrained', None)
+
+    model.save_pretrained = save_pretrained
+    model._seq_cls_arch_patched = True
```
hjh0119, line 327: Two blockers before merge: the wrapper is a bare local closure on the instance capturing the original model/config (patcher.py:286/290/312), so copy.deepcopy(model).save_pretrained() writes the original weights silently and pickle.dumps(model) raises — both fine on main. Please hook the model class (origin = cls.save_pretrained, gate on an instance flag) or use MethodType + functools.wraps, reading
state from self; and replace == [class_name] with a pre-save snapshot comparison (that also fixes the architectures=None case)

RerankerGuo: Both points addressed in c121140.

**Class-level hook.** The wrapper is now installed on the model class (`cls.save_pretrained`) and gated by the `_seq_cls_arch_patched` instance flag, so nothing is stored on the instance anymore. Consequences:

- `copy.deepcopy(model).save_pretrained()` now saves the copy's own weights — the copy carries the flag and the wrapper resolves `self`/`self.config` at call time instead of holding the original objects in a closure. `test_seq_cls_architectures_saved_config_deepcopy` fills the copy's `score` weight with a sentinel, saves, reloads with `AutoModelForSequenceClassification`, and asserts the sentinel came back.
- `pickle.dumps(model)` round-trips again — the closure was never picklable. `test_seq_cls_architectures_saved_config_pickle` covers the round trip plus a save from the unpickled model.
- The `_seq_cls_arch_wrapped` marker on the wrapper keeps a second patched model of the same class from double-wrapping, and `test_seq_cls_architectures_unpatched_instance_unaffected` pins the gate: an unpatched `Qwen2ForCausalLM` saves byte-identically to before.

**Pre-save snapshot instead of `== [class_name]`.** The architectures value is snapshotted before the save starts, and the config hook restores it only when the save changed it. That subsumes the class-name special case and fixes two things the `==` check got wrong:

- a value assigned after patching is now actually kept (the old code restored the patch-time value, clobbering the later assignment — `test_seq_cls_architectures_post_patch_assignment_kept`);
- a config whose `architectures` is `None` no longer gets `null` written into `config.json`; with nothing to restore, transformers' own `[class_name]` is left in place (`test_seq_cls_architectures_none_keeps_class_name`).

All four cases fail by construction on baaf078 (verified by re-running against it). `pytest tests/general/test_model.py` → 9 passed; the 2 failures are the pre-existing network-dependent `test_qwen2`/`test_modelscope_hub`. `yapf`/`isort`/`flake8` clean.

PR #10002, opened 2026-08-28
perf(npu): vectorize ring attention LSE extractio

Ruihan11:
# PR type
- [ ] Bug Fix
- [ ] New Feature
- [ ] Document Updates
- [ ] More Models or Datasets Support
- [x] Optimization

# PR information

## Summary

Vectorize second-half LSE extraction in the Ascend NPU ring-attention backward path.

## Motivation

`_get_second_half_lse` previously iterated over every packed sequence and called `.item()` for each sequence boundary. When `cu_seqlens` is an NPU tensor, these per-sequence scalar reads introduce repeated NPU-to-CPU synchronizations.

The overhead grows with the number of packed sequences and becomes significant for padding-free batches containing many short sequences.

## Changes

- Replace per-sequence slicing and copying with device-side index construction.
- Use `repeat_interleave` to build source and destination offsets.
- Use a single `index_select` to gather the second-half LSE values.
- Preserve the existing function signature, output layout, dtype, device, and autograd behavior.
- Retain a bounded-loop fast path for up to four packed sequences to avoid regressions for single long sequences.
- Keep the CUDA implementation, ring schedule, HCCL communication, and native NPU attention APIs unchanged.

## Tests

Added:

- CPU reference-based correctness tests.
- Ascend NPU correctness tests for:
  - `float32`
  - `bfloat16`
  - variable-length packed sequences
  - empty segments
  - `int32` and `int64` `cu_seqlens`
  - supported LSE layouts
  - backward gradient propagation
- Opt-in NPU microbenchmark.
- Opt-in 2-rank HCCL end-to-end ring-attention test.

Validation:

```text
22 passed, 2 skipped
python -m compileall -q swift: passed
flake8: passed
isort: passed
yapf: passed
git diff --check: passed
2-rank HCCL E2E: passed
```

## Experiment results

Environment:

```text
Ascend 910B3
torch==2.10.0+cpu
torch_npu==2.10.0
bfloat16
8 attention heads
total length: 4096 tokens
warmup: 10
measurement: 30
```

Helper-level benchmark:

| Packed sequences | Tokens/sequence | Baseline | Optimized | Speedup |
|---:|---:|---:|---:|---:|
| 1 | 4096 | 229.3 us | 229.6 us | 1.00x |
| 8 | 512 | 1072.8 us | 853.4 us | 1.26x |
| 32 | 128 | 3930.1 us | 864.4 us | 4.55x |
| 128 | 32 | 15265.7 us | 874.0 us | 17.47x |
[...]

PR #10064, opened 2026-09-08
perf(cuda): vectorize ring attention LSE extractio

Ruihan11:
# PR type
- [ ] Bug Fix
- [ ] New Feature
- [ ] Document Updates
- [x] Optimization

# PR information

## Summary

Follow-up to #10002.

Vectorize CUDA LSE extraction in the packed-sequence ring-attention backward path to avoid repeated device-to-host synchronization.

## Motivation

`get_half_lse` previously iterated over every packed sequence, read each CUDA sequence boundary with `.item()`, and performed per-sequence slice assignments. This introduced one or more host/device synchronization points per packed sequence and became increasingly expensive for padding-free batches containing many short sequences.

## Changes

- Construct packed-sequence indices on the CUDA device with `repeat_interleave` and `arange`.
- Gather the selected LSE values with one `index_select` operation.
- Retain the existing bounded-loop path for fewer than eight packed sequences, where vectorized index construction is not beneficial.
- Preserve the `@torch.jit.script` decorator, function signature, front/back semantics, output layout, dtype, device, autograd behavior, and valid sequence-length divisibility contract.
- Leave the NPU implementation and ring schedule unchanged.

## Tests

- CPU and CUDA correctness against an independent loop reference.
- FP32 and BF16 inputs.
- `int32` and `int64` `cu_seqlens`.
- Front and back selection.
- Variable-length and empty packed segments.
- Forward outputs and backward gradients.
- Threshold coverage for 1/2/4/7/8/16 packed sequences.
- Real 2-rank NCCL E2E using `zigzag_ring_flash_attn_varlen_func` and FlashAttention forward/backward.

Validation completed:

```text
Focused pytest: 46 passed, 1 skipped
2-rank NCCL CUDA E2E: passed
compileall: passed
pre-commit: passed
flake8, isort, yapf: passed
git diff --check: passed
```

## Experiment results

Measured on 8x NVIDIA A800-SXM4-80GB with PyTorch 2.10.0+cu130, CUDA 13.0, and FlashAttention 2.8.3. Results use 20 warmup iterations and 100 synchronized measurement iterations.

Median p50 latency across the benchmark matrix:

| Packed sequences | Baseline | Optimized | Speedup |
|---:|---:|---:|---:|
| 1 | 62.9 us | 63.8 us | 0.99x |
| 2 | 100.6 us | 101.6 us | 0.99x |
| 4 | 173.3 us | 174.0 us | 0.99x |
| 8 | 226.8 us | 151.2 us | 1.50x |
| 16 | 424.8 us | 151.0 us | 2.81x |
| 32 | 820.7 us | 150.8 us | 5.46x |
| 64 | 1611.8 us | 151.2 us | 10.66x |
| 128 | 3185.9 us | 150.8 us | 21.10x |
| 256 | 6332.1 us | 151.2 us | 41.87x |

Across the complete benchmark matrix:

- No p50 or p95 case regressed by more than 5%.
- The performance crossover is at 8 packed sequences.
- For 8 or more sequences, the minimum p50 speedup is 1.46x and the minimum p95 speedup is 1.42x.
- The maximum peak allocation increase is 128 KiB.

Nsight Systems profile for 16,384 tokens and 256 packed sequences:

| Metric | Baseline | Optimized |
|---|---:|---:|
| `cudaStreamSynchronize` calls | 8,865 | 1 |
| Device-to-host scalar copies | 8,864 | 0 |
| CUDA kernel launches | 7,686 | 428 |

The bounded-loop path remains unchanged for small packed batches, while the vectorized path removes the per-sequence synchronization and scalar device-to-host copies for larger batches.

## Reproduction

Run the helper correctness and threshold tests from the repository root:

```bash
pytest -q tests/sequence_parallel/test_zigzag_ring_attn.py
```

Run the 2-rank distributed CUDA E2E test. It is opt-in and needs two visible GPUs plus `flash-attn`:

```bash
SWIFT_RUN_CUDA_E2E=1 CUDA_VISIBLE_DEVICES=0,1 \
    torchrun --standalone --nproc_per_node=2 \
    -m pytest -q -s tests/sequence_parallel/test_zigzag_ring_attn_cuda_e2e.py
```

Expected output. The per-rank losses differ because each rank seeds from its own local rank:

```text
rank=0 output=(1, 36, 2, 8) loss=0.510180
rank=1 output=(1, 36, 2, 8) loss=0.374627
1 passed
```

Without `SWIFT_RUN_CUDA_E2E=1` or a distributed launcher the E2E test skips, which is why a plain `pytest` run reports one skip.

The latency numbers above come from a standalone harness that calls `get_half_lse` directly and compares the current implementation against the previous loop, using identical inputs in separate processes, `torch.cuda.synchronize()` around each timed call, 20 warmup and 100 measured iterations, packed sequence counts from 1 to 256, 4,096 and 16,384 total tokens, FP32 and BF16, and both front and back selection. The harness is not part of this PR; I can include it if that helps review.

## Limitations

- This change targets CUDA packed-sequence LSE extraction in the ring-attention backward path only.
- Existing valid sequence-length divisibility requirements are unchanged.
- Inputs violating those requirements remain unsupported.
- The NPU implementation is intentionally unchanged.

### DIFF
--- a/swift/sequence_parallel/zigzag_ring_attn.py
+++ b/swift/sequence_parallel/zigzag_ring_attn.py
@@ -50,6 +50,23 @@ def get_half_lse(lse, cu_seqlens, *, front: bool):
     Returns:
         The filtered lse with the same shape as lse
     """
+    # Vector index construction has fixed overhead; use it only for sufficiently packed inputs.
+    _LSE_VEC_PATH_MIN_NUMEL = torch.jit.annotate(int, 9)
+    if cu_seqlens.numel() >= _LSE_VEC_PATH_MIN_NUMEL:
+        cu_seqlens = cu_seqlens.to(device=lse.device, dtype=torch.long)
+        lengths = cu_seqlens[1:] - cu_seqlens[:-1]
+        half_lengths = lengths // 2
+        half_total_len = lse.shape[1] // 2
+        if front:
+            source_starts = cu_seqlens[:-1]
+        else:
+            source_starts = cu_seqlens[:-1] + half_lengths
+        destination_starts = cu_seqlens[:-1] // 2
+        source_indices = torch.repeat_interleave(source_starts, half_lengths, output_size=half_total_len)
+        destination_indices = torch.repeat_interleave(destination_starts, half_lengths, output_size=half_total_len)
+        source_indices = source_indices + torch.arange(half_total_len, device=lse.device) - destination_indices
+        return lse.index_select(1, source_indices)
+
     new_lse = torch.empty(
         (lse.shape[0], lse.shape[1] // 2),
         dtype=lse.dtype,
--- a/tests/sequence_parallel/test_zigzag_ring_attn.py
+++ b/tests/sequence_parallel/test_zigzag_ring_attn.py
@@ -1,0 +1,76 @@
+import pytest
+import torch
+
+from swift.sequence_parallel.zigzag_ring_attn import get_half_lse
+
+
+def _make_cu_seqlens(lengths, dtype=torch.int32, device='cpu'):
+    lengths = torch.tensor(lengths, dtype=dtype, device=device)
+    return torch.cat((lengths.new_zeros(1), lengths.cumsum(0)))
+
+
+def _reference_get_half_lse(lse, cu_seqlens, front):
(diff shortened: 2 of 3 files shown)


hjh0119 (approved):

hjh0119:
thanks!

PR #10070, opened 2026-09-08
fix(embedding): select UEmbed dense token with vLLM

taking-lying-flat:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

UEmbed's vLLM embedding path uses sequence pooling (`embed` / LAST), returning the final appended sparse EOS state instead of the dense state at `-(num_eos_tokens + 1)`. Its original checkpoint also stores `language_model.*` / `visual.*` keys without the CausalLM `model.` prefix.

Configure UEmbed for `token_embed` / ALL with activation disabled, then select the configured dense position and L2-normalize the single-vector response. A small, lazily registered Qwen3.5 subclass accepts both original and already-prefixed checkpoints without rewriting weight files. Registration uses `vllm.general_plugins` so it also works in spawned workers. Editable installations need reinstalling to pick up the new entry point.

The synchronous embedding path also needs `add_request` and embedding response conversion; it previously called the asynchronous `encode` API and attempted to format a chat completion. Other embedding models retain their sequence-pooling configuration and returned vectors.

Reference: [official UEmbed vLLM example](https://github.com/Alibaba-NLP/UEmbed/blob/main/examples/vllm_example.py). The existing 16-token template suffix is preserved.

## Experiment results

- Local regression checks (test scripts retained outside this PR): **9 passed, 2 skipped**. Covers offsets 0/3/16, unequal sequence lengths, synchronous/asynchronous responses, unchanged non-UEmbed vectors, and loading original/exported keys through the real vLLM pooling adapter and weight loader with a tiny CPU parameter tree. The two full-checkpoint GPU alignment checks were skipped; cosine similarity > 0.999 remains unverified.
- Additional checks passed: architecture resolution and key mapping in a real spawned process; the public `infer()` path for UEmbed and ordinary embedding models in both modes with a simulated backend, including reversed completion order.
- `uvx pre-commit run --from-ref upstream/main --to-ref HEAD`: all applicable hooks passed. `git diff --check` passed.

### DIFF
--- a/swift/infer_engine/vllm_engine.py
+++ b/swift/infer_engine/vllm_engine.py
@@ -221,6 +221,9 @@ class VllmEngine(InferEngine):
                 ignore_patterns=getattr(template.model_meta, 'ignore_patterns', None),
                 hub_token=hub_token)
         super().__init__(template)
+        self._uembed_num_eos_tokens = (
+            getattr(template, 'num_eos_tokens', 0)
+            if self.task_type == 'embedding' and self.model_meta.model_type == 'qwen3_5_emb' else None)
         if max_model_len is not None:
             self.max_model_len = max_model_len
             logger.info(f'Setting max_model_len: {max_model_len}')
@@ -348,6 +351,13 @@ class VllmEngine(InferEngine):
         arch_mapping = {'deepseek_vl2': ['DeepseekVLV2ForCausalLM'], 'chatglm4v': ['GLM4VForCausalLM']}
         if self.model_meta.model_type in arch_mapping:
             hf_overrides['architectures'] = arch_mapping[self.model_meta.model_type]
+        if self._uembed_num_eos_tokens is not None:
+            from vllm.config import PoolerConfig
+
+            from .patch import register_uembed_model
+            register_uembed_model()
+            hf_overrides['architectures'] = ['UEmbedForConditionalGeneration']
+            engine_kwargs['pooler_config'] = PoolerConfig(task='token_embed', pooling_type='ALL')
         hf_overrides.update(self._get_hf_config_overrides())
         if hf_overrides:
             engine_kwargs['hf_overrides'] = hf_overrides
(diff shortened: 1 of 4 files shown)


swift/infer_engine/vllm_engine.py
```diff
@@ -348,6 +351,13 @@ def _prepare_engine_kwargs(self, max_model_len, engine_kwargs) -> None:
         arch_mapping = {'deepseek_vl2': ['DeepseekVLV2ForCausalLM'], 'chatglm4v': ['GLM4VForCausalLM']}
         if self.model_meta.model_type in arch_mapping:
             hf_overrides['architectures'] = arch_mapping[self.model_meta.model_type]
+        if self._uembed_num_eos_tokens is not None:
+            from vllm.config import PoolerConfig
+
+            from .patch import register_uembed_model
+            register_uembed_model()
+            hf_overrides['architectures'] = ['UEmbedForConditionalGeneration']
+            engine_kwargs['pooler_config'] = PoolerConfig(task='token_embed', pooling_type='ALL')
```
hjh0119, line 360: 特定模型逻辑放到对应的template prepare_engine_kwargs 中

swift/infer_engine/vllm_engine.py
```diff
@@ -497,11 +507,15 @@ def _add_request(self,
                 pooling_kwargs = {}
                 if has_task_arg:
                     pooling_kwargs['task'] = task_mapping[self.task_type]
+                if self._uembed_num_eos_tokens is not None:
+                    pooling_kwargs.update(task='token_embed', use_activation=False)
                 if self.task_type in ('reranker', 'generative_reranker') and \
                         has_activation_arg and self.reranker_use_activation:
                     pooling_kwargs['activation'] = True
                 pooling_params = PoolingParams(**pooling_kwargs)
```
hjh0119, line 515: 确认下use_activation和activation的区别以及兼容性

swift/infer_engine/vllm_engine.py
```diff
@@ -696,7 +710,11 @@ def _format_prompt_logprobs(prompt_logprobs):
 
     def _create_embedding_response(self, result, generation_config, request_id) -> EmbeddingResponse:
         assert result is not None
-        embedding = result.outputs.data.cpu().numpy().tolist()
+        embedding = result.outputs.data
+        if self._uembed_num_eos_tokens is not None:
+            embedding = torch.nn.functional.normalize(
+                embedding[-(self._uembed_num_eos_tokens + 1)].float(), p=2, dim=-1)
+        embedding = embedding.cpu().tolist()
```
hjh0119, line 717: 放到template里

z0o0ey:
Thank you very much for your discovery and code contribution. We have resolved this issue based on your code [#10074 ](https://github.com/modelscope/ms-swift/pull/10074) please test it to see if the problem is resolved.

taking-lying-flat:
> Thank you very much for your discovery and code contribution. We have resolved this issue based on your code [#10074 ](https://github.com/modelscope/ms-swift/pull/10074) please test it to see if the problem is resolved.

OK。thanks

PR #10083, opened 2026-09-09
fix(dataset): detect failed streaming packing workers

MrCapricornLiu:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

With streaming packing and `strict=True`, an encoding error other than
`MaxLengthError` terminates a packing worker. The iterator still waits for one
queue result per submitted sample, so it blocks indefinitely. An unexpected
worker exit has the same effect.

Poll the output queue and check the workers' process sentinels when no result
arrives. Raise a `RuntimeError` pointing to the worker logs if a worker has
exited. Sentinels also work when a forked DataLoader process inherits the
packing workers; checking `exitcode` there cannot detect processes owned by
its parent. Live workers may take longer than the polling interval, and
non-strict errors and `MaxLengthError` retain their existing skip behavior.

## Experiment results

```bash
CUDA_VISIBLE_DEVICES='' OMP_NUM_THREADS=1 TOKENIZERS_PARALLELISM=false \
  HF_HUB_OFFLINE=1 python -m pytest \
  tests/general/test_packing_multiprocessing_context.py -q
```

All 40 tests pass, including seven regressions for strict errors, abrupt exit,
skipped samples, slow workers, and failure/live-worker behavior through a
forked DataLoader consumer. The two original worker-failure regressions fail
without the fix. Changed-file pre-commit and diff checks pass.

Tested on Linux with Python 3.12 and real spawned packing workers. Full model
training, distributed training and Windows/macOS were not run. Workers that
remain alive but are stuck inside an encoder are outside this change.

### DIFF
--- a/swift/dataset/packing.py
+++ b/swift/dataset/packing.py
@@ -3,6 +3,8 @@ import math
 import multiprocessing as mp
 import torch.distributed as dist
 from itertools import chain
+from multiprocessing.connection import wait
+from queue import Empty
 from torch.utils.data import Dataset, IterableDataset
 from tqdm import tqdm
 from typing import Optional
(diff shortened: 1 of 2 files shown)


Excelius-Wang (approved):
Using process sentinels makes sense for the forked DataLoader case, where the consumer isn't the packing worker's parent.

I ran all seven new worker-error tests on Linux, including real spawn workers. Strict encoding errors and abrupt exits reach the consumer, while slow workers, non-strict errors and `MaxLengthError` keep their intended behavior. I also checked failure detection directly and through a DataLoader: reverting the queue change makes both checks hit their deadline instead of raising.

I didn't find a blocking issue in this path. I haven't tested Windows/macOS or distributed training; a live encoder that never returns remains outside this fix.

tastelikefeet (approved):

PR #10109, opened 2026-09-11
[Bugfix][RLHF] fix colocate offload and vLLM wake-up ordering

rxlqn:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Summary

- Reload only models that were actually moved to CPU by `offload_context`.
- Avoid overlapping trainer and colocated vLLM weights during same-step resampling.
- Preserve the required wake/sync order when a new training step needs a weight update.
- Add regression coverage for selective reload and both vLLM wake-up paths.

## Root cause

`prepare_rollout()` can run before `accelerator.prepare()`. In that phase a DeepSpeed ZeRO-2 actor may still be entirely CPU-resident. The old context unconditionally reloaded it on exit even though nothing had been offloaded, placing the full model on one GPU and causing an initialization OOM. This is the same lifecycle class described in #9793, but it also affects DeepSpeed ZeRO-2.

There was a second ordering issue during dynamic resampling: when the global step had not changed, vLLM weights were woken before entering `offload_context`, even though no weight sync was needed. That temporarily kept both trainer and vLLM weights on GPU.

## Changes

- Make `offload_model()` return whether any parameter or buffer was moved to CPU.
- Track and reload only models actually offloaded by the current context.
- When weights need synchronization, wake vLLM weights and sync before offloading the trainer, then wake the KV cache inside the context.
- When no synchronization is needed, offload the trainer before waking vLLM weights and the KV cache together.
- Add unit tests for CPU-resident initialization, selective actor/ref reload, same-step resampling, and new-step weight synchronization.

## Experiment results

- `pre-commit run --all-files` — all hooks passed (`flake8`, `isort`, `yapf`, whitespace/YAML/EOF/requirements/quotes/conflict/LF checks).
- `PYTHONPATH=. python3 -m unittest tests.utils.test_rollout_offload_order -v` — 9 tests passed.
- `python3 -m py_compile swift/rlhf_trainers/rollout_mixin.py swift/rlhf_trainers/grpo_trainer.py tests/utils/test_rollout_offload_order.py` — passed.
- 9B multimodal GRPO smoke test with DeepSpeed ZeRO-2 optimizer offload, colocated vLLM, `offload_model=true`, and `offload_optimizer=true` — completed 3/3 training steps, covering initialization, same-step resampling, optimizer updates, and subsequent weight synchronization.

## Related

- Related to #9793.


hjh0119 (approved):
thanks for your contribution, can you also modify the megatron side?

rxlqn:
Thanks for pointing this out. I’ve updated the Megatron colocate rollout path in commit `f35f72fc5`.

The Megatron path now follows the same memory-safe lifecycle:

- For a new training step, wake vLLM weights, synchronize the updated weights, offload the trainer, and then wake the KV cache.
- For same-step resampling, offload the trainer before waking vLLM weights and the KV cache.
- vLLM versions without tagged `wake_up()` are also covered.

Validation:

- Added 3 Megatron-specific rollout ordering regression tests.
- 12 related Transformers/Megatron unit tests passed.
- Ran a 2-GPU Megatron GRPO colocate smoke test with Qwen3.5-2B, `sleep_level=2`, `offload_model=true`, and `offload_optimizer=true`; all 3/3 training iterations completed successfully without OOM.
- Pre-commit checks passed for the changed files.

hjh0119 (approved):

hjh0119:
thanks, can you pass the lint test?

hjh0119:
The related lint fixes have been addressed in https://github.com/modelscope/ms-swift/pull/10153. Ready to merge

PR #10111, opened 2026-09-11
Add optional batch shuffling after length grouping

GodRayyy:
# PR type

- [ ] Bug Fix
- [x] New Feature
- [x] Document Updates

# PR information

With `group_by_length=True`, each mega-batch is sorted from long to short. This creates repeated length cycles. The Gemma4 examples already note that limited shuffling can cause loss fluctuations.

Add `group_by_length_shuffle_batches`, disabled by default. It shuffles complete DP micro-batch blocks after length grouping and before rank sharding. Each local batch stays intact, so padding and rank assignments are preserved. The first block stays first for early OOM detection. An incomplete tail stays last.

This covers both the standard and Megatron length-grouped samplers. Resume skips are applied after the deterministic permutation. Keep the option unchanged when resuming. It does not change samples within a block or guarantee better convergence.

## Experiment results

Gemma4-E4B-it LoRA, 2 RTX 4090 GPUs, SDPA, batch size 2 per GPU, gradient accumulation 2. Each run used the same 256 synthetic text samples for 32 optimizer steps. Encoded lengths ranged from 39 to 407 tokens.

| Sampling | Padding slots | Adjacent batch length correlation | Valid tokens/s |
| --- | --- | --- | --- |
| Random | 29.65% | 0.075 | 1838 |
| Length grouped | 3.34% | 0.761 | 1978 |
| Length grouped + block shuffle | 3.34% | 0.132 | 1970 |

The new option reduced length correlation and preserved padding efficiency. Throughput was similar. These are short smoke tests, not convergence results.

Validation: `pre-commit run --all-files` passed. The focused sampler, resume-seed, and import tests passed (15 tests). Sampler tests cover 1, 2, 4, 8, and 56 simulated DP ranks, tails, TP mapping, epochs, and resume skips. They pass with Transformers 5.3.0 and 5.5.4 for the new sampler tests. Full Megatron training was not run.

A separate Gemma4 run resumed from step 16 to 32 with matching per-rank batch metrics. Final weights were not bitwise identical; the unmodified sampler showed the same limitation.


GodRayyy:
I am closing this PR for now. The tests confirm that batch shuffling preserves padding and reduces length-order correlation. They do not yet show a consistent training benefit. I will revisit this after larger-scale testing.

PR #10116, opened 2026-09-12
fix(rlhf): undo gradient-accumulation scaling for DAPO/CISPO/FIPO losses

yhb643:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

## Problem

`dapo`, `cispo` and `fipo` normalize the policy loss with the completion tokens of the **whole
gradient-accumulation window**:

```python
elif self.loss_type in ['cispo', 'dapo', 'fipo']:
    # CISPO, DAPO, and FIPO: Normalize by total completion tokens across all processes
    normalizer = grpo_batch.num_items_in_batch / self.accelerator.num_processes
    loss = (per_token_loss * completion_mask).sum() / normalizer
```

`num_items_in_batch` is built from `ga_batch_encoded_inputs` and `_gather_and_flatten(...)`
(grpo_trainer.py:803-809), so it already covers every micro-batch of the accumulation window on
every process. A single micro-batch therefore contributes only a **slice** of the global token
mean.

`Trainer.training_step` then divides the loss by `current_gradient_accumulation_steps` before
backward (transformers forces `num_steps=1` on the `Accelerator` plugin, so the loss is not
scaled a second time there). For a loss that averages over the current micro-batch that division
is correct, but for these three losses it shrinks the loss and its gradient by the number of
accumulation steps:

| | gradient of one update step |
| --- | --- |
| intended (global token mean) | `g` |
| actual, `gradient_accumulation_steps=4` | `g / 4` |

The logged loss is off in the same way: it is divided by the number of accumulation steps
instead of being the global token mean.

### Minimal reproduction (CPU, no model)

```python
loss = torch.tensor(2.0, requires_grad=True)                    # window-normalized loss of one micro-batch
scaled = loss * trainer.current_gradient_accumulation_steps     # patch
(scaled / trainer.current_gradient_accumulation_steps).backward()
# before the patch the gradient of every micro-batch is 1 / gradient_accumulation_steps smaller
```

## Fix

Compensate the window-normalized policy loss right after the normalizer, inside the
`cispo`/`dapo`/`fipo` branch, so the Trainer division cancels out before the auxiliary losses are
added:

```python
elif self.loss_type in WINDOW_NORMALIZED_LOSS_TYPES:
    normalizer = grpo_batch.num_items_in_batch / self.accelerator.num_processes
    loss = (per_token_loss * completion_mask).sum() / normalizer
    loss = self._undo_gradient_accumulation_scaling(loss)
```

* only `cispo`/`dapo`/`fipo` (`WINDOW_NORMALIZED_LOSS_TYPES`); `grpo`/`sapo`/`bnpo`/`dr_grpo` are
  averaged over the current micro-batch, so the Trainer scaling is already correct for them;
* only where the Trainer itself scales (`training`, `not model_accepts_loss_kwargs`,
  `compute_loss_func is None`), mirroring its condition;
* `getattr(self, 'current_gradient_accumulation_steps', self.args.gradient_accumulation_steps)`
  keeps the fix working on older transformers versions, where the attribute does not exist yet;
* applying it *before* SDAR/CHORD leaves those micro-batch means with the Trainer division they
  need (an earlier revision multiplied the combined loss, which made them
  `gradient_accumulation_steps` times too strong), and the reported policy loss becomes the global
  token mean instead of being divided by the accumulation steps.

The Liger path (`use_liger_loss`) is not touched, it has its own normalization in
`liger_grpo_loss`.

## Experiment results

`tests/test_align/test_grpo_loss.py` runs on CPU only, no model / NPU / GPU required:

```text
$ pytest -q tests/test_align/test_grpo_loss.py
25 passed
```

It checks that

* the gradient after the Trainer division is independent of `gradient_accumulation_steps`
  (1, 2, 4, 8) for `dapo`/`cispo`/`fipo`;
* the accumulated SDAR/CHORD gradient stays the window mean for 1/2/4/8 accumulation steps (this
  fails on the previous placement from 2 steps up);
* the fallback to `args.gradient_accumulation_steps` works on transformers versions without
  `current_gradient_accumulation_steps`;
* `grpo`/`sapo`/`bnpo`/`dr_grpo` are not modified;
* no compensation happens during evaluation, when the model accepts loss kwargs, or when
  `compute_loss_func` is set.

### DIFF
--- a/swift/rlhf_trainers/grpo_trainer.py
+++ b/swift/rlhf_trainers/grpo_trainer.py
@@ -77,6 +77,10 @@ if is_wandb_available():
 if is_swanlab_available():
     import swanlab
 
+# Losses normalized by the completion tokens of the whole gradient-accumulation window instead of the tokens of
+# the current micro-batch. See `GRPOTrainer._undo_gradient_accumulation_scaling`.
+WINDOW_NORMALIZED_LOSS_TYPES = ['cispo', 'dapo', 'fipo']
+
 
 class GRPOTrainer(RolloutTrainerMixin, SwiftMixin, HFGRPOTrainer):
     executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
@@ -1128,7 +1132,7 @@ class GRPOTrainer(RolloutTrainerMixin, SwiftMixin, HFGRPOTrainer):
             if self.beta != 0.0:
                 kl_loss = (per_token_kl * completion_mask).sum() / completion_mask.sum().clamp(min=1.0)
                 loss = loss + kl_loss * self.beta
-        elif self.loss_type in ['cispo', 'dapo', 'fipo']:
+        elif self.loss_type in WINDOW_NORMALIZED_LOSS_TYPES:
             # CISPO, DAPO, and FIPO: Normalize by total completion tokens across all processes
             normalizer = grpo_batch.num_items_in_batch / self.accelerator.num_processes
             loss = (per_token_loss * completion_mask).sum() / normalizer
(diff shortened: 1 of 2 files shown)


Excelius-Wang (changes requested):
Could the GAS compensation happen before adding SDAR/CHORD? Those auxiliary losses are micro-batch means, so they still need the Trainer's division by GAS. Multiplying the combined loss cancels that division for them too.

I reproduced this on CPU by executing `_compute_loss_and_metrics` and `compute_sdar_loss` with a minimal trainer stub: DAPO, four identical tokens, zero advantages, student log-prob -1, teacher log-prob -0.5, SDAR coefficient 0.1. After accumulating `loss / GAS`, the gradient is -0.092414 at GAS=1 but -0.369657 at GAS=4. Reverting this PR keeps both at -0.092414.

Please compensate the window-normalized loss before the auxiliary terms and add an auxiliary-gradient regression test. I haven't run full training.

yhb643:
@Excelius-Wang you're right — thanks for the CPU repro. Fixed in 79e89fd9.

The compensation now happens inside the window-normalized branch, right after the normalizer and before SDAR/CHORD, so those keep the Trainer's division by the accumulation steps. Your numbers match: multiplying the combined loss left the auxiliary gradient at `contribution * GAS` instead of the window mean.

Added `test_auxiliary_loss_keeps_trainer_gradient_accumulation_scaling`, which pins the accumulated auxiliary gradient to the window mean for 1/2/4/8 accumulation steps — it fails on the previous placement from 2 steps up. The test file now passes 25 cases.

My PR description's claim that the auxiliary terms were "scaled up and divided down again, so unchanged" was wrong: they were unchanged relative to the pre-fix behaviour, but that behaviour is the correct one for a micro-batch mean. I've corrected the description.

hjh0119 (approved):

PR #10118, opened 2026-09-12
docs: fix custom reward function examples

BingH225:
# PR type
- [ ] Bug Fix
- [ ] New Feature
- [x] Document Updates

# PR information

Fix the custom reward function examples in both the Chinese and English GRPO developer guides:
- Add the missing colon to the DummyLengthRewardFunction class declaration.
- Add self to the three synchronous __call__ examples so they work as instance methods.

The length-reward example currently raises SyntaxError. Fixing only the colon still leaves a TypeError when the instance receives completions by keyword.

## Experiment results

- Reproduced both failures from the original snippets.
- Parsed the corrected snippets and tested the length-reward example with the ORM base class extracted from the repository, without importing the training stack. Completion lengths 0, 1024, and 1025 return [0.0, 0.0, 1.0] with both positional and keyword calls.
- Executed both dataset-column method examples with a lightweight trainer-state fixture.
- git diff --check and Markdown parsing passed.

The configured pre-commit hooks passed for both changed files.

No training or model downloads were performed; runtime code is unchanged.

### DIFF
--- a/docs/source/Instruction/GRPO/DeveloperGuide/reward_function.md
+++ b/docs/source/Instruction/GRPO/DeveloperGuide/reward_function.md
@@ -9,8 +9,8 @@
 
 ```python
 from swift.rewards import ORM, orms
-class DummyLengthRewardFunction(ORM)
-    def __call__(completions, **kwargs):
+class DummyLengthRewardFunction(ORM):
+    def __call__(self, completions, **kwargs):
         return [1.0 if len(completion) > 1024 else 0.0 for completion in completions]
 
 orms['dummy']= DummyLengthRewardFunction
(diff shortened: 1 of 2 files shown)


Excelius-Wang (approved):
Both fixes are needed here: the colon makes the class parse, and `self` makes the examples callable as reward instances.

I ran the three changed examples from both guides with the real ORM base class. The length example returns `[0.0, 0.0, 1.0]` for completion lengths 0, 1024 and 1025 with either positional or keyword arguments, and both dataset-column examples accept `trainer_state` correctly. The original snippets fail these checks; the updated ones pass. No training run was needed for this documentation change.

tastelikefeet (approved):

PR #10119, opened 2026-09-13
fix(dataset): list hub IDs for named datasets in the Web UI

aprylewu:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

The dataset dropdowns read registry keys as `(ms_id, hf_id, path)` tuples, but `register_dataset` also stores entries under a string `dataset_name`. The built-in `self-cognition` therefore appears as `s` with ModelScope or `e` with Hugging Face. A one-character custom name also raises `IndexError` in HF mode.

Read the selected hub ID from each `DatasetMeta` instead. This restores the actual dataset IDs while preserving order and filtering out entries without an ID for that hub. The training, sampling and export dropdowns all use this helper.

## Experiment results

Using normal `swift.dataset` imports with Python 3.10 and locally cached dependencies in offline mode:
- `tests/general/test_dataset_list.py`: 2 tests with both hub modes pass. On main, the same tests fail with corrupted IDs and an `IndexError`.
- `tests/general/test_dataset_source.py`: 2 tests pass.
- `pre-commit run --all-files`: passed; the new test file also passes all applicable hooks.

No model downloads or training runs were needed. Developed with Codex assistance.

### DIFF
--- a/swift/dataset/register.py
+++ b/swift/dataset/register.py
@@ -13,13 +13,11 @@ logger = get_logger()
 
 def get_dataset_list():
     datasets = []
-    for key in DATASET_MAPPING:
-        if use_hf_hub():
-            if key[1]:
-                datasets.append(key[1])
-        else:
-            if key[0]:
-                datasets.append(key[0])
+    use_hf = use_hf_hub()
+    for dataset_meta in DATASET_MAPPING.values():
+        dataset_id = dataset_meta.hf_dataset_id if use_hf else dataset_meta.ms_dataset_id
+        if dataset_id:
+            datasets.append(dataset_id)
     return datasets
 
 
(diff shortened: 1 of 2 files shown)


Excelius-Wang (approved):
Looks good to me. Registry keys can be names or tuples, so reading the hub ID from DatasetMeta avoids treating a dataset name as a sequence.

I reproduced the broken self-cognition entry and the single-character name crash with the old code. The new tests pass in both hub modes, including entries with no ID for the selected hub, and the dataset-source tests pass too. I checked the training, sampling and export dropdown call sites, but didn't exercise the browser UI.

tastelikefeet (approved):

PR #10120, opened 2026-09-13
fix(dataset): preserve JSON-serialized messages with boolean and null fields

aprylewu:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

JSON-serialized messages containing `true`, `false`, or `null` currently fail in `default_repair_messages()`, which only uses `ast.literal_eval()`. For example, a CSV message containing `"loss": false` raises `ValueError` with `strict=True` and is dropped with the default filtering behavior. Serialized OpenAI tool calls with `"content": null` fail before normal tool-call conversion.

Parse JSON first, then fall back to the existing Python literal parser. Existing single-quoted Python representations and already parsed message lists retain their behavior. The same fix applies to serialized `rejected_messages`.

## Experiment results

- New regression suite: **3 passed / 4 failed before**, **7 passed after**.
- Existing provider/rejected-message and dataset-source tests: **9 passed**.
- `pre-commit run --all-files`: passed.
- `git diff --check`: passed.

The regressions use temporary CSV/JSONL files through the real `load_dataset()` and Arrow preprocessing paths, including a default-mode case where a valid row was previously dropped. They also cover Python-literal compatibility, native lists, malformed input, and rejected messages. Tests use normal project imports; no model downloads or training runs were needed.

Developed and validated with OpenAI Codex assistance.

### DIFF
--- a/swift/dataset/preprocessor/core.py
+++ b/swift/dataset/preprocessor/core.py
@@ -438,7 +438,10 @@ class AlpacaPreprocessor(ResponsePreprocessor):
 
 def default_repair_messages(s: Union[str, Any]) -> Any:
     if isinstance(s, str):
-        return ast.literal_eval(s)
+        try:
+            return json.loads(s)
+        except json.JSONDecodeError:
+            return ast.literal_eval(s)
     return s
 
 
--- a/tests/general/test_serialized_messages.py
+++ b/tests/general/test_serialized_messages.py
@@ -1,0 +1,124 @@
+# Copyright (c) ModelScope Contributors. All rights reserved.
+import csv
+import json
+import os
+import tempfile
+import unittest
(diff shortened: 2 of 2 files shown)


Excelius-Wang (approved):
Looks good to me. Parsing JSON before normalizing tool calls handles `content: null` correctly, and falling back to `ast.literal_eval` preserves support for existing Python-literal inputs.

I reproduced the valid-row drop with the old parser and confirmed the new regression tests pass with this change. I also checked serialized `rejected_messages` through CSV loading, streaming JSONL, and filtering a malformed row alongside a valid one; all passed.

My validation was limited to CPU dataset preprocessing; I didn't run training.

tastelikefeet (approved):

Issue #6695, opened 2025-11-21
设置MODELSCOPE_CACHE后会报Stale file handle


gakkiri:
[rank1]:   File "/usr/local/lib/python3.10/dist-packages/datasets/arrow_dataset.py", line 4101, in _select_with_indices_mapping
[rank1]:     os.chmod(indices_cache_file_name, 0o666 & ~umask)
[rank1]: OSError: [Errno 116] Stale file handle: xxx

每个node都会报相同错误

slin000111:
> [rank1]: File "/usr/local/lib/python3.10/dist-packages/datasets/arrow_dataset.py", line 4101, in _select_with_indices_mapping [rank1]: os.chmod(indices_cache_file_name, 0o666 & ~umask) [rank1]: OSError: [Errno 116] Stale file handle: xxx
> 
> 每个node都会报相同错误

这个钉钉群里问一下？“魔搭ModelScope开发者联盟群 ①”群的钉钉群号： 44837352

mingwu2333:
遇到了一样的问题，请问解决了吗

tiggerwu:
Same problem. Have you solved the problem? Thank.

CoderChen01:
Same problem.

tomzw11:
filelock在释放锁后会删除锁文件，升级filelock==3.32.0后解决了

https://github.com/tox-dev/filelock/pull/577

PR #10121, opened 2026-09-13
fix(dataset): tag datasets before post_process to avoid cross-rank cache race

zmazak0:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

`_inject_dataset_routing_tag` (added in #9678) calls `Dataset.add_column`, which falls back to
`flatten_indices()` whenever the dataset carries an indices table. Sampling (`dataset#n`),
`dataset_shuffle` and `split_dataset_ratio` all go through `select()`, so an indices table is
present in ordinary training runs.

`flatten_indices()` is a `map()`, and `map()` derives its cache file path from the dataset
fingerprint (`Dataset._get_cache_file_path`): `<dataset dir>/cache-<fingerprint>.arrow`. The
fingerprint is deterministic, so **every rank computes the same path**. When the dataset directory
lives on a filesystem shared between nodes — the normal setup for multi-node training — all ranks
write that same file concurrently and race on the `shutil.move()` → `os.chmod()` at the end of
`Dataset._map_single`. The losing ranks die:

```
File ".../swift/dataset/loader.py", line 221, in _inject_dataset_routing_tag
  return dataset.add_column('dataset', [ds_name] * len(dataset))
File ".../datasets/arrow_dataset.py", line 6241, in add_column
  dataset = self.flatten_indices() if self._indices is not None else self
File ".../datasets/arrow_dataset.py", line 4048, in _map_single
  os.chmod(cache_file_name, 0o666 & ~umask)
FileNotFoundError: [Errno 2] No such file or directory: '.../cache-c1bda3f0ecec9469.arrow'
```

A rank that loses the race one line earlier fails instead with
`Failed to open local file '.../cache-<fingerprint>.arrow'`.

Setting `load_from_cache_file=False` does not help: `flatten_indices()` does not accept that
argument and always goes through `map()`'s regular cache path.

This is the only place in the dataset loading path without such protection — `RowPreprocessor.__call__`
already guards its `map()` calls with `safe_ddp_context` plus the
`if not load_from_cache_file and is_dist() and not is_master()` fallback, and `loader.py` itself uses
`safe_ddp_context` for loading and for `train_test_split`.

## Fix

The routing tag is constant per source dataset, so the order of tagging and sampling/splitting does
not matter. This PR moves the injection to right after `loader.load()`, while the dataset still has
no indices table. `add_column()` then stays on the in-memory path: no cache file is written, there is
nothing for the ranks to race on, and the flatten pass over every dataset disappears (a measurable
win on its own — it was ~1.5 min per 360k-row dataset, on every rank). The val dataset inherits the
column from the train dataset it is split from, so the separate injection for it is no longer needed.

An alternative would be to wrap the existing call in `safe_ddp_context`, which is more idiomatic for
this repo. I did not take that route because it serializes the flatten onto the master rank: with a
job that lists ~90 datasets the remaining ranks would sit on a barrier for a long time. Happy to
switch if you prefer that shape.

## Experiment results

Added `tests/general/test_dataset_routing_tag.py` — hermetic, no model or network access; it
registers a temporary local jsonl dataset and asserts that `flatten_indices()` is never called while
the routing tag is applied.

```
# with this patch
Ran 2 tests in 1.143s
OK

# on main, without the loader.py change
FAIL: test_routing_tag_is_added_without_flattening
AssertionError: Expected 'flatten_indices' to not have been called. Called 1 times.
Ran 2 tests in 1.052s
FAILED (failures=1)
```

Originally hit on a 4-node / 32-rank SFT run with ~90 datasets on a shared mount, where it killed the
job during dataset loading. Related, still open: #6695 reports the same class of failure
(`os.chmod` on a shared cache path from every rank) in `_select_with_indices_mapping`.


tastelikefeet (approved):

Excelius-Wang (approved):
Moving the constant routing tag before sampling and splitting preserves its meaning and avoids the extra flattening in the covered load path.

I reproduced the flatten_indices call with the old ordering; the new regression passes without it. I also checked validation-only loading and a streaming train/validation split, and both retain the correct source tag. These checks use local JSONL files and the real dataset loader. I haven't reproduced the multi-node shared-filesystem race itself or measured the performance improvement.

PR #10133, opened 2026-09-14
fix(infer): propagate streaming generation thread errors

Excelius-Wang:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

If model.generate raises during streaming, its inner thread exits without notifying TokensIteratorStreamer. The outer inference worker stays blocked waiting for another token, so both the failing request and subsequent valid requests hang. For example, repetition_penalty=0 raises a ValueError promptly in non-streaming mode but strands the worker in streaming mode.

Send generation exceptions through the existing streamer queue and raise them in the consumer thread. This lets the existing worker handler notify batched callers and continue serving requests. Preserve partial tokens, normal completion and the existing strict-mode policy.

## Experiment results

9 tests and 10 subtests pass, covering failures before/after partial output, batched callers, logprobs, recovery, strict/non-strict behavior and existing streaming error regressions. With an actual tiny Qwen2 and message records from three public datasets, the original ValueError now reaches callers and subsequent valid non-stream/stream requests complete, with and without logprobs. Changed-file pre-commit checks pass. No GPU OOM or full HTTP deployment test was run.


tastelikefeet (approved):

tastelikefeet:
Please lint by `pre-commit run --all-files`

PR #10159, opened 2026-09-16
fix(dataset): replace existing source routing columns

MrCapricornLiu:
# PR type
- [x] Bug Fix

# PR information

Loading a dataset that already has a `dataset` column fails with "duplicated
columns" when `remove_unused_columns=False`. The routing-tag helper appends
another column with the same name. Its streaming branch already replaces that
field with the current source name.

Remove the previous column before adding the routing tag on map-style datasets.
This preserves the other columns and keeps the behavior consistent with streaming.
The tag is still added before sampling and splitting, retaining #10121's cache
handling.

## Experiment results

```bash
python -m unittest tests.general.test_dataset_routing_tag tests.general.test_dataset_source -v
```

All six tests pass using local JSONL files and the real dataset loader. The two
new tests fail on the original code. Coverage includes streaming/non-streaming,
column pruning on/off, sampled train/validation splits, unchanged input datasets,
and the existing no-flattening check. No models or network access are needed.

Changed-file pre-commit passes; the all-files check retains the unrelated
formatting failure in `tests/megatron/test_infonce_ddp_e2e.py` from #10150.


hjh0119:
plz pass the lint test

hjh0119 (approved):

hjh0119:
The related lint fixes have been addressed in https://github.com/modelscope/ms-swift/pull/10153. Ready to merge

PR #10161, opened 2026-09-16
fix(train): serialize GaLore projector state in checkpoints

MrCapricornLiu:
GaLore stores a `GaLoreProjector` instance in optimizer state. Consequently, a normal optimizer checkpoint cannot be restored by Trainer's `torch.load(..., weights_only=True)`: it fails with an unsupported-global error before training resumes.

Serialize the projector as a dictionary of configuration values and tensors, then reconstruct it when loading optimizer state. The live optimizer keeps its projector object. Move cached projection bases to the gradient's device/dtype before reuse, so CPU-mapped checkpoints can resume on CUDA without waiting for the next SVD refresh. This applies to the bundled AdamW, Adafactor and AdamW8bit implementations.

Validation:
- Four regression methods pass on CUDA, including five projection modes on CPU, CPU-mapped restoration on CUDA, bitsandbytes 8-bit state, ordinary parameter groups and legacy in-memory projector states.
- A tiny GPT-2 with the real Transformers Trainer fails to resume on main. With the patch, resuming at step 3 produces exactly the same parameters as uninterrupted training through step 6.
- Changed-file pre-commit hooks and `git diff --check` pass. Base-revision whole-repository lint has an unrelated formatting failure in `tests/megatron/test_infonce_ddp_e2e.py`.

This makes newly written checkpoints compatible with safe loading. It does not change safe-loading policy for older pickled-object checkpoints, or address the separate per-parameter optimizer wrapper's checkpoint handling. Validation used FP32 parameters; distributed/sharded checkpoints and q-galore were not tested.


tastelikefeet (approved):

tastelikefeet:
lint the code by:
```
pre-commit run --all-files
```
please

PR #10163, opened 2026-09-16
fix(reward): parse complete generative reward scores

MrCapricornLiu:
`GenRMPlugin.extract_reward` reads `Reward: 10` as 1, reads `Reward: 1e-1` as 1, and cannot parse negative scores. This can change reward values before `compute_rewards` averages them.

Parse signed decimal and scientific-notation scores without imposing a fixed scoring range. Malformed or non-finite values return `None`, preserving the existing behavior of excluding invalid choices and returning zero when every choice is invalid. The default prompt still requests scores in [0, 1], while custom scoring scales remain supported. Markdown delimiters and sentence-ending punctuation continue to work.

Five regression tests pass, covering custom scoring scales, CJK suffixes, scientific notation, invalid/non-finite values, and aggregation using `ChatCompletionResponse` objects. Changed-file pre-commit checks pass. Validation covers parsing and aggregation locally; it does not call a reward model or run GRPO training.


swift/rewards/rm_plugin.py
```diff
@@ -119,25 +119,24 @@ def prepare_rm_inputs(self, inputs: List[Dict]) -> List[Dict]:
...
+        # Parse the whole number before checking its range, rather than accepting a 0/1 prefix.
+        match = re.search(r'Reward:\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:[eE][+-]?\d+)?)(?!\w|\.\d)', model_output)
         if match:
-            return float(match.group(1))
-        else:
-            logger.warning("Unable to extract reward score from the model's output, set reward to 0")
-            return None
+            reward = float(match.group(1))
+            if 0 <= reward <= 1:
+                return reward
```
hjh0119: Do we need to tie the range here?

MrCapricornLiu: No. The [0, 1] check came from the default example prompt, but it shouldn't constrain the parser when the scoring scale is customized. I'll remove that bound and keep full numeric parsing plus rejection of non-finite values, with regression tests for negative scores, scores above 1, and scientific notation.

hjh0119 (approved):

swift/rewards/rm_plugin.py
```diff
@@ -119,25 +120,24 @@ def prepare_rm_inputs(self, inputs: List[Dict]) -> List[Dict]:
...
         Returns:
-            float: The extracted reward score.
-
-        Raises:
-            ValueError: If the reward score cannot be extracted or the format is incorrect.
+            The extracted finite score, or None if the score is invalid.
         """
-        match = re.search(r'Reward:\s*([0-1](?:\.\d+)?)', model_output)
+        # Parse the whole number instead of accepting a prefix of a multi-digit or exponent score.
+        match = re.search(r'Reward:\s*([+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:[eE][+-]?\d+)?)(?!\w|\.\d)', model_output)
```
hjh0119: One regression to fix before merge: (?!\w) is Unicode-aware, so Reward: 0.85分 / Reward: 10分 now return None (both parsed fine before).

Please switch the boundary to (?![0-9A-Za-z_]|\.\d) and add a CJK-suffix test case.

MrCapricornLiu: Updated the boundary to `(?![0-9A-Za-z_]|\.\d)` and added CJK-suffix cases for decimal, integer, negative and scientific-notation scores. All five tests pass, including the existing malformed-number checks.

PR #10190, opened 2026-09-17
Fix preserving output_router_logits in multimodal template forward hooks

### LINKED ISSUE #10189: [Bug] Qwen3.5 MoE multimodal DPO drops output_router_logits and crashes with aux_loss=None
I have searched the existing issues and did not find a report for this multimodal native DPO parameter-passthrough failure.

## Environment

- ms-swift: 4.5.2 (affected local checkout)
- transformers / trl / torch: exact versions are available from the original environment but are omitted from this sanitized report
- Python: 3.10.21
- CUDA: 12.9
- GPU: NVIDIA H100 80GB
- Model: Qwen3.6-35B-A3B (Qwen3.5 MoE architecture)
- Training: native Swift DPO/RLHF + LoRA + DeepSpeed
- Template: Qwen3.5 multimodal template

## Configuration

The relevant option is:

```text
--router_aux_loss_coef 1e-3
```

When this coefficient is positive, the DPO trainer sets:

```python
batch['output_router_logits'] = True
```

## Symptom

The model initializes successfully and NCCL setup completes, but the first training step fails with:

```text
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```

The failure occurs in the DPO trainer at:

```python
losses = losses + self.aux_loss_coef * model_output['aux_loss']
```

`model_output['aux_loss']` is `None`.

## Root cause

The Qwen3.5 multimodal template reuses the Qwen2VL `_post_encode` path. That path returns a new kwargs dictionary containing `inputs_embeds`. `Template.pre_forward_hook` then restores selected values from the original kwargs using an explicit whitelist.

The whitelist does not contain `output_router_logits`, so the flag set by the DPO trainer is dropped before the Qwen3.5 MoE model forward. The model consequently uses its default `output_router_logits=False`, leaves `aux_loss=None`, and the DPO trainer multiplies a float by `None`.

The effective call path is:

```text
DPO sets output_router_logits=True
  -> multimodal _post_encode rebuilds kwargs
  -> pre_forward_hook drops output_router_logits
  -> Qwen3.5 MoE computes no aux_loss
  -> DPO crashes on float * None
```

This is distinct from NCCL initialization and from independent CUDA/ABI loading issues that may appear later during distributed shutdown.

## Minimal fix

Add `output_router_logits` to the kwargs restoration whitelist in `Template.pre_forward_hook`:

```python
'output_router_logits',
```

This preserves the caller-provided flag and allows the model to compute the router auxiliary loss. It does not change the DPO loss formula or convert `aux_loss=None` to zero.

## Validation

With the one-line change, the same DPO run using `--router_aux_loss_coef 1e-3` completes the first training step and reports a numeric auxiliary loss, for example:

```text
'nll_loss': '0.2953', 'aux_loss': '8.07'
```

I am preparing a small regression test that simulates a post-encode path rebuilding kwargs and verifies that `output_router_logits=True` reaches the model forward.

## Questions

1. Should `output_router_logits` be added to the generic `Template.pre_forward_hook` restoration whitelist?
2. Should a regression test cover a multimodal Qwen3.5 MoE DPO/SFT batch with `router_aux_loss_coef > 0`?
3. Should the hook preserve model-forward kwargs more generically instead of maintaining a manually curated whitelist?

hkxxxxx:
## Summary

Preserve `output_router_logits` when `Template.pre_forward_hook` restores model-forward kwargs after multimodal `_post_encode`.

Fixes #10189.

## Background

Native Swift DPO/SFT can enable the MoE router auxiliary loss with:

```python
output_router_logits = True
```

The Qwen3.5 multimodal template reuses a Qwen2VL post-encoding path that returns a new kwargs dictionary. The generic template hook restores only fields in a manually maintained whitelist. Because `output_router_logits` was missing from that whitelist, the flag was silently removed before the Hugging Face model forward.

With `router_aux_loss_coef > 0`, Qwen3.5 MoE then returns `aux_loss=None`; native DPO fails when it evaluates:

```python
self.aux_loss_coef * model_output['aux_loss']
```

## Changes

- Add `output_router_logits` to the forward-kwargs restoration whitelist in `Template.pre_forward_hook`.
- Add a regression test covering a post-encode path that rebuilds kwargs while preserving `output_router_logits=True`.

## Validation

- Reproduced the failure with Qwen3.5/Qwen3.6 MoE native DPO, LoRA, DeepSpeed, and `router_aux_loss_coef=1e-3`.
- Confirmed the first Python failure is `aux_loss=None`, not NCCL initialization.
- Confirmed the one-line fix produces a numeric `aux_loss` and allows the training step to continue.
- `conda run -n llm python -m pytest tests/general/test_template_forward_hook.py -q` — `1 passed`.
- `python -m py_compile swift/template/base.py tests/general/test_template_forward_hook.py`.
- `git diff --check`.

## Compatibility and risk

This change only preserves a caller-provided forward kwarg. It does not change the default model behavior when the kwarg is absent, and it does not alter the auxiliary-loss coefficient or loss formula.

The field is already used by the native DPO/SFT trainers for supported MoE models. The regression test targets the parameter-loss mechanism directly and does not require model weights or private data.

## Checklist

- [x] Bug Fix
- [x] Add/update regression test.
- [x] Run the relevant focused test.
- [x] Confirm no private paths, logs, credentials, or local environment files are included.
- [ ] Run a full Qwen3.5 MoE smoke test when the target hardware/environment is available.


tastelikefeet:
Conflicting

tastelikefeet (approved):

PR #10202, opened 2026-09-19
Fix MiniCPM-o bounds offset after left padding

zm249:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

MiniCPM-o image and audio bounds are computed before batch collation.

During batched inference, shorter samples are left-padded by the base
collator, but `image_bound` and `audio_bounds` were not shifted by the
corresponding padding length. This causes multimodal bounds to become
misaligned for shorter samples when `batch_size > 1`.

This PR shifts `image_bound` and `audio_bounds` by each sample's actual
left-padding length after the base collator determines the padded sequence
length.

The adjustment is applied only when padding-free mode is disabled and the
effective padding side is left.

# Experiment results

- Batch size 1: behavior is unchanged.
- Batch size 4: bounds of shorter samples receive the correct padding offset.
- The longest sample remains unchanged because its padding length is zero.
- Local collator checks passed.
- Existing related tests passed.


swift/template/templates/minicpm.py
```diff
@@ -617,6 +618,24 @@ def _data_collator(self, batch: List[Dict[str, Any]], *, padding_to: Optional[in
         res['audio_bounds'] = audio_bounds_list if audio_bounds_list else []
 
         res.update(Template._data_collator(self, batch, padding_to=padding_to))
+
+        # image_bound and audio_bounds are computed before batching. Keep them aligned with input_ids when the
+        # base collator left-pads shorter samples. In inference, the base collator always uses left padding;
```
tastelikefeet: Thanks for solving this issue! I think other minicpm templates share the same issue, this function can be used as a common helper all of them

zm249: Thanks for pointing this out! You're right — MiniCPMVTemplate (including v2.5/v2.6) and MiniCPMV4_5Template use the same pre-padding image_bound handling, while MiniCPMO4_5 additionally has audio_bounds.

I'll extract the padding-offset logic into a shared helper and apply it to all affected MiniCPM collators. The helper will account for the effective padding side and skip the adjustment for padding-free or right-padding cases. I'll update the PR accordingly.

zm249: Thanks for the suggestion. I've extracted the padding offset logic into a shared helper and applied it to all affected MiniCPM templates. MiniCPM-o 4.5 uses the same helper for both image_bound and audio_bounds. Right-padding and padding-free cases remain unchanged.

tastelikefeet (approved):

PR #10207, opened 2026-09-20
fix(rollout): acknowledge IPC buckets after all TP copies

### LINKED ISSUE #10204: [Bug] Potential shared-buffer overwrite race in Ray Megatron weight sync: ACK is sent before the TP barrier
### Checklist / 检查清单

### Bug Description / Bug 描述

While investigating corrupted first-rollout output in Ray + Megatron GRPO, I noticed a possible shared-buffer lifetime race in the native IPC weight synchronization path.

**Scope of this report:** the ACK/barrier ordering below is present in upstream source. We have observed rollout corruption in our environment, but have **not yet established that this race is the root cause of that corruption**, nor validated the proposed fix end to end.

#### Receiver acknowledges one rank before all TP ranks finish reading

Checked against upstream `main` at commit `654e24f17b5d9f40ed4b9ee4c56a723320244db5`.

In [WeightSyncWorkerExtension.update_weights_from_ipc](https://github.com/modelscope/ms-swift/blob/654e24f17b5d9f40ed4b9ee4c56a723320244db5/swift/pipelines/infer/rollout.py#L485-L522), each TP worker copies tensors out of the shared buffer, then executes:

```python
if _torch.cuda.is_available():
    _torch.cuda.synchronize()

if is_driver:
    socket.send(b'')  # bucket received

if tp_size > 1:
    _dist.barrier(group=cpu_group)
```

On the [sender side](https://github.com/modelscope/ms-swift/blob/654e24f17b5d9f40ed4b9ee4c56a723320244db5/swift/ray/megatron/rollout/weight_transfer.py#L100-L139), `_flush()` waits for this ACK, resets the offset, and subsequent `_process()` calls overwrite the same `self.buffer`.

This appears to permit the following interleaving:

1. All TP ranks receive metadata for bucket A.
2. TP rank 0 finishes its local copies and sends the ACK.
3. The sender receives the ACK and starts writing bucket B into the same shared buffer.
4. A slower non-driver TP rank is still reading/copying bucket A.
5. The TP barrier completes only afterward; it cannot undo the overwrite.

The local CUDA/HIP synchronization does not wait for copy completion in other TP worker processes. The sender is not a participant in the TP barrier. Consequently, tensor names/shapes could remain valid while copied values are corrupted, without a synchronization exception.

Expected invariant: **the sender must not reuse a bucket until every receiving TP rank has finished reading it.**

#### Environment and observed symptom

- Local source checkout: ms-swift `4.6.0.dev0`, based on upstream `d3c1a795a20d359476659450792de4cadf889fda`, with platform/model-loading integration changes. The receiver ACK/barrier ordering is unchanged from upstream; it is also present in the current upstream commit linked above.
- Linux, AMD MI308X / ROCm.
- Model: Qwen3.6-27B, `model_type=qwen3_5`, BF16; text-only math GRPO.
- Ray cluster: 4 nodes × 8 GPUs.
- Training: 16 GPUs, TP=4, PP=2, DP=2.
- Rollout: 16 GPUs, four vLLM replicas, TP=4 per replica.
- Native initial weight synchronization enabled; GPU IPC path.
- Reusable transfer bucket: 3072 MiB.
- mcore-bridge 1.6.3; Transformers 5.9.0.

Observed with both tested ROCm software stacks:

| PyTorch | vLLM | Initial vLLM load_format | First rollout after initial synchronization |
| --- | --- | --- | --- |
| 2.10.0 | 0.17.0 | dummy and auto, tested separately | Corrupted/repetitive text |
| 2.13.0 | 0.27.0 | dummy and auto, tested separately | Corrupted/repetitive text |

The first generation occurs **before the first optimizer update**. Initial weight synchronization returns successfully. Changing `load_format` from `dummy` to `auto` does not skip the subsequent initial synchronization, and did not resolve the symptom.

For each `auto` run, we inspected the first 256 completions (16 prompts × 16 samples):

- vLLM 0.17.0: all 256 received zero accuracy reward; 175 contained Unicode replacement characters.
- vLLM 0.27.0: all 256 received zero accuracy reward; 204 contained Unicode replacement characters.

The issue is visible in the generated text, not just in reward parsing: outputs contain long repeated words such as `sequence` / `Step`, symbols, and malformed text.

Standalone vLLM 0.27.0 loading the same checkpoint directly produced coherent correct responses for a matching test prompt. However, that standalone environment used Transformers 5.17 instead of 5.9, so this is **supporting evidence, not a fully controlled pre/post-sync comparison**. ROLL also works with the same model/data, but uses a different synchronization implementation and is not an otherwise identical experiment.

#### Focused regression test to isolate the suspected race

This is a proposed test; we have not run this instrumented GPU test yet:

1. Exercise `BucketedWeightSender` + `update_weights_from_ipc` with TP >= 2 and at least two known-value buckets.
2. Delay one non-driver TP rank immediately before it copies a bucket, allowing the driver to finish first.
3. Check the private copied tensors on every rank against the announced bucket's expected values.
4. Verify that the sender cannot start refilling the shared buffer until the delayed rank's copy has completed.
5. Repeat after changing the ACK ordering, and test GPU IPC (and the CPU shared-memory path where applicable).

This should help separate a transport lifetime race from model conversion or vLLM weight-reload issues.

A possible minimal correction is to acknowledge the bucket **after** all TP ranks have synchronized their local copies and reached the TP barrier:

```python
if _torch.cuda.is_available():
    _torch.cuda.synchronize()

if tp_size > 1:
    _dist.barrier(group=cpu_group)

if is_driver:
    socket.send(b'')  # all TP ranks have copied this bucket
```

This is a proposed fix for review, **not a confirmed fix for our end-to-end rollout corruption**. A per-rank completion protocol could enforce the same buffer-lifetime guarantee.

Related reports:
- #9787 describes Qwen3.5 rollout degradation after synchronization, but uses ZeRO3 + external vLLM, not this Ray + Megatron IPC path. I am not assuming the same root cause.
- #9627 fixes a separate ROCm visible-device/IPC issue; that change is already included in our checkout.

Is there an additional cross-process buffer-lifetime guarantee that makes the current ACK-before-barrier ordering safe? Otherwise, would moving the ACK after the barrier, together with a delayed-rank regression test, be the appropriate fix?

taking-lying-flat:
# PR type

- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

Fixes #10204.

Credit to @JeremySun1224 for identifying and analyzing the race in #10204 and proposing the ACK-after-TP-barrier correction implemented here.

When the TP driver finishes copying a weight bucket before another TP rank, its ACK lets `BucketedWeightSender` overwrite the shared buffer while that rank is still reading it. Move the bucket ACK after the TP barrier, keeping local CUDA synchronization before the barrier, so all receiving ranks finish copying before the sender can reuse the buffer.

Also request an explicit copy in the SHM path: `.to(device)` can alias the shared buffer when the receiver is on CPU, leaving weights vulnerable to reuse even after the barrier.

## Experiment results

- Local CPU protocol regression: 16 cases covering two known-value buckets, TP=1/2, cached IPC and real SHM mappings, full-weight and LoRA loads, and empty final buckets. A simulated sender overwrites the buffer immediately on ACK; a simulated delayed peer copies at the barrier. Before the fix, the isolated receiver produced 12 tensor-value failures; after the fix, all 16 cases passed with the receiver module imported normally and unrelated vLLM engine imports stubbed on CPU.
- Existing `tests/train/test_vllm_weight_sync.py`: 4 passed.
- `pre-commit run --all-files` and `git diff --cached --check`: passed.

The local regression harness is not included in this PR. Actual CUDA/ROCm IPC and end-to-end Ray + Megatron GRPO were not run. This fixes the buffer-lifetime race; whether it explains the rollout corruption reported in the issue still needs validation on the affected cluster.


taking-lying-flat:
Thank you, @JeremySun1224, for identifying and documenting the shared-buffer lifetime race in #10204, and for proposing the ACK-after-TP-barrier fix. Both the diagnosis and the core ordering correction in this PR originate from your report.

Your explanation of the sender/receiver interleaving, the buffer-lifetime invariant, and the delayed-rank regression strategy provided the basis for this implementation and its local validation. I also appreciate the care you took to distinguish the race from the still-unconfirmed cause of the observed rollout corruption. That distinction is preserved in this PR.

Thank you for the careful investigation, the concrete proposed correction, and the time you invested in making this issue actionable.

tastelikefeet (approved):

PR #10211, opened 2026-09-20
fix(deploy): reject media URLs that resolve to non-public addresses

### LINKED ISSUE #9740: Unauthenticated SSRF via multimodal image_url/audio/video in swift deploy (no URL filtering)

RerankerGuo:
## What this fixes

Fixes #9740 — unauthenticated SSRF in the OpenAI-compatible `swift deploy` server.

The deployment server fetches `image_url` / `audio_url` / `video_url` server-side, binds `0.0.0.0` and has `api_key=None` by default, so any client can make the server issue requests to internal-only services and the cloud instance-metadata endpoint. Confirmed in the report: `load_file("http://127.0.0.1:8456/secret.png")` returns the bytes, and a redirect is followed, so a public host can 302 to `169.254.169.254`.

`load_file` is the single choke point for images (`load_image`), audio (`_load_audio_librosa`) and video (`_resolve_video_local_path`), so the guard lives there:

- **Resolve and check every address.** The hostname is resolved with `getaddrinfo` and the URL is rejected if *any* returned address is not globally routable (`ipaddress.is_global`). This covers loopback, RFC1918, link-local / `169.254.169.254`, IPv6 ULA/loopback, and IP-literal bypasses (`2130706433`, `127.1`, `0.0.0.0`). Because the *resolved* addresses are checked, DNS names that point into private space (`localtest.me`, `nip.io`, …) are rejected too.
- **Validate every redirect hop.** `session.get(..., allow_redirects=False)` plus a bounded manual follow loop, re-validating each `Location`. This closes the redirect bypass the report calls out, and also rejects non-http schemes in a redirect target.
- **No fetcher fallback.** `_load_audio_librosa` re-raises `UnsafeMediaURLError` instead of falling back to `audioread`/ffmpeg, which would otherwise retry a blocked URL through a different code path.

## Compatibility

- Public URLs are unaffected: the media URLs already used by the test suite (`modelscope-open.oss-cn-hangzhou.aliyuncs.com`, `example.com`) resolve to global addresses and still load.
- Deployments that legitimately serve media from an internal address can opt out with `SWIFT_ALLOW_PRIVATE_MEDIA_URLS=1` (same `SWIFT_*` env-var convention as `SWIFT_TIMEOUT`), so internal/self-hosted image services keep working.
- Non-URL inputs (local paths, base64, data URIs, bytes) never reach the guard.

## Tests

New `tests/general/test_vision_utils.py`:

- `test_ip_literal_urls` — table of IP-literal URLs (metadata, loopback in all notations, RFC1918, IPv6, non-http scheme vs. public) asserting allow/block. IP literals only, so no DNS is needed and the cases are deterministic.
- `test_hostname_resolving_to_private_address` — a public hostname that resolves to `127.0.0.1` is rejected.
- `test_load_file_rejects_loopback_before_any_request` — `load_file` refuses before any request is made.
- `test_load_file_validates_every_redirect_hop` — a public host that 302s to the metadata endpoint is blocked at the hop.
- `test_load_file_follows_safe_redirect` — a redirect that stays on public addresses is still followed, with `allow_redirects=False` asserted on each call.
- `test_audio_fallback_does_not_bypass_guard` — the librosa fallback does not retry a blocked URL.
- `test_opt_out_env_var` — `SWIFT_ALLOW_PRIVATE_MEDIA_URLS=1` restores the previous behaviour.

Verified locally with a real loopback HTTP server (reporter's PoC): before the change the server was hit and the bytes returned; after, `UnsafeMediaURLError` is raised and the server is never contacted, for the direct fetch, the redirect and the audio fallback.

## Checklist

- [x] `pre-commit run --all-files` passes locally (yapf / flake8 / isort clean)
- [x] `tests/general/test_vision_utils.py` passes (7 tests)
- [x] No regression in `tests/general/test_template*.py`, `test_arch.py`, `test_optional_template_dependencies.py`


tastelikefeet:
Previous PR has protected the deploy command from this security issue, in fact, we think training and inference does not need this, because the developer self must keep it clean:
#10062「Fix downloading security」:https://github.com/modelscope/ms-swift/pull/10062
#10066「Fix security issue」:https://github.com/modelscope/ms-swift/pull/10066

PR #10214, opened 2026-09-21
fix(train): avoid autograd graphs in fully frozen vision towers

linjunxiong:
# PR type
- [x] Bug Fix
- [ ] New Feature
- [ ] Document Updates

# PR information

A fully frozen vision encoder can still retain backward activations when input-gradient hooks make its embedding output differentiable. For Qwen3.5 with Transformers 5.2.0, `model.enable_input_require_grads()` installs hooks on nested models and owns their handles on the parent; disabling input gradients on the vision submodel does not remove those parent-owned hooks. This can cause OOM for image-heavy SFT batches even with the ViT and aligner frozen.

Wrap registered vision-tower forwards during trainer preparation. Run without autograd only when **all parameters are frozen and no input needs gradients**. Recheck these conditions on each forward, preserving trainable aligners/adapters, partially frozen towers, dynamic unfreezing, and gradients to upstream inputs. Unknown input containers conservatively keep the existing behavior. Hooks remain installed so that later unfreezing still works, and the wrapper respects an enclosing `no_grad` context.

Related to #9945, which reduces activation memory through ViT checkpointing. This change avoids retaining an unnecessary graph in the fully frozen case. It is separate from #10142's shared trainability-cache fix for partially frozen layers.

## Experiment results

Automated regression tests use CPU, PyTorch 2.6.0+cu124 and Transformers 5.2.0 with the complete ms-swift source checkout:

```bash
python -m unittest discover -s tests/general -p test_frozen_vision.py -v
```

- A synthetic frozen encoder reproduces the input-grad-hook issue and saves zero tensors for backward after patching, with identical output.
- Feature/head and encoder gradients match the eager reference for a trainable aligner, partially unfrozen tower, real PEFT LoRA adapter, and positional/nested differentiable inputs.
- Repeated patching, copying a patched module, unfreezing/refreezing, and an enclosing `no_grad` context are covered.
- A randomly initialized tiny Qwen3.5 exercises the actual trainer preparation method and full image/text forward/backward. Logits, loss and every parameter gradient match an eager reference with ViT checkpointing disabled/enabled and reentrant/non-reentrant checkpointing.

Historical measurement motivating this PR: on 8 H100 GPUs, Qwen3.5-9B full language-model SFT with a frozen ViT/aligner, 32K, ZeRO-3 and SP=8, a 32570-token batch containing 55 images failed with OOM. A Qwen3.5-specific `no_grad` workaround reduced PyTorch allocated memory immediately before `lm_head` from 65.46 GiB to 20.64 GiB for the same batch. A four-update replay (two initial optimizer updates followed by the failing batch and its successor) passed. These are point-in-time allocations, **not total GPU peaks**, and were measured with ms-swift 4.1.1; the generalized patch in this PR has not repeated that distributed benchmark.

All four test methods pass. Disabling the new wrapper reproduces four assertion failures: retained backward tensors and differentiable frozen-vision outputs in each of the three Qwen3.5 checkpointing modes. All repository pre-commit hooks pass (`pre-commit run --all-files`).

### Transformers 5.12.1 compatibility (d9579df)

Reproduced the review concern: all three Qwen3.5 checkpoint-mode subtests raised `ValueError` because their manually constructed multimodal inputs lacked `mm_token_type_ids`, now required for M-RoPE. The synthetic frozen-encoder tests already passed.

The test now supplies image token types (1 for image tokens, 0 for text) when the model forward explicitly supports this argument. This preserves Transformers 5.2.0 compatibility and exercises the normal multimodal position-ID path; it does not skip the test or weaken its gradient assertions. The production fix is unchanged.

The full `test_frozen_vision.py` suite passes (4 test methods) on **both Transformers 5.2.0 and 5.12.1**, on CPU with PyTorch 2.6.0+cu124 and PEFT 0.18.1. Pre-commit checks for the changed file also pass.


tests/general/test_frozen_vision.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertIsNone(tower.aligner.weight.grad)
+        self.assertFalse(tower(x).requires_grad)
+        tower.aligner.requires_grad_(True)
+        self.assertTrue(tower(x).requires_grad)
+        with torch.no_grad():
+            self.assertFalse(tower(x).requires_grad)
+        tower.requires_grad_(False)
+        self.assertFalse(tower(x).requires_grad)
+
+    def test_qwen35_trainer_preparation(self):
```
tastelikefeet, line 116: maybe error in transformers 5.12.1

linjunxiong: Thanks for catching this. Reproduced on Transformers 5.12.1: the test inputs were missing mm_token_type_ids. Fixed in d9579df by supplying it when supported. All four tests now pass on both 5.2.0 and 5.12.1. The production fix is unchanged.

tastelikefeet (approved):

PR #10222, opened 2026-09-21
fix(infer): align TransformersEngine with doc defaults by propagating…

DreamBinary:
### Summary
Fixes the issue where `TransformersEngine` lacked the `strict` attribute and did not receive `args.strict` during pipeline initialization. This PR properly passes `args.strict` from `SwiftInfer` down to `TransformersEngine` and sets its default value to `False` in compliance with the official documentation.

---

### Reference Documentation
According to the [Swift Command-line Parameters Documentation](https://swift.readthedocs.io/en/latest/Instruction/Command-line-parameters.html):

> **strict**: If True, any malformed row in the dataset will raise an error; otherwise, erroneous samples are dropped. **Default is False.** This is typically used for debugging.

---

### Bug & Root Cause
When running inference with `TransformersEngine`, inspect/debug outputs showed that `strict` was never initialized on the engine instance:

```bash
CUDA_VISIBLE_DEVICES=0 swift infer \
    --model weights/Qwen3.5-0.8B \
    --stream true \
    --infer_backend transformers \
    --max_new_tokens 2048
> /opt/qwenvl/lib/python3.12/site-packages/swift/infer_engine/transformers_engine.py(124)__init__()->None
-> breakpoint()
(Pdb) print(getattr(self, 'strict'))
*** AttributeError: 'TransformersEngine' object has no attribute 'strict'


hjh0119:
The premise needs re-checking: strict is intentionally read via getattr(self, 'strict', True) — the True fallback was deliberately flipped from False in #3203, whose same commit added explicit strict = False to vanilla_sampler / distill_sampler / rewards-prm (i.e. "engine defaults to raising, data-gen paths opt out"). Making the constructor default False and wiring args.strict (False by default) therefore silently flips the default behaviour of swift infer / eval / RL rollout / prm-orm from "raise" to "return an Exception placeholder in the result list", yet none of those consumers filters such entries (only vanilla_sampler.py:160 does). Verified end-to-end: the real AssertionError: response_role: "nope" degrades into AttributeError: 'AssertionError' object has no attribute 'choices', hiding the root cause (parse_data_from_response even returns None silently). Suggest either dropping the second hunk (or using strict: Optional[bool] = None for back-compat), or completing the drop semantics in SwiftInfer / eval / rollout and wiring it consistently for all backends. Also note the cited doc is the dataset-level strict, which already reaches load_dataset via get_dataset_kwargs() — not the same thing as engine-level template.encode failures.

PR #10224, opened 2026-09-21
feat: add Moore Threads MUSA backend support via torchada

caizhi-mt:
Run ms-swift on Moore Threads GPUs (torch_musa) through torchada, which redirects the torch.cuda.* APIs to torch.musa, instead of forking every CUDA code path.

- Import torchada at the top of `swift/__init__.py` when torch_musa and torchada are both installed, so the patches are active before anything touches torch.cuda. Nothing changes on other platforms.
- torchada intentionally leaves `torch.cuda.is_available()` False on MUSA, so the device helpers can no longer rely on it. Add `is_torch_musa_available()` and MUSA branches to the helpers in `swift/utils/torch_utils.py` (device, set_device, device_count, empty_cache, ipc_collect, synchronize, get_torch_device) and default the process group backend to `mccl`.
- Use `musa:<rank>` for the default device map and check `torch.musa.is_bf16_supported()` when choosing the default dtype.
- `select_device` also sets MUSA_VISIBLE_DEVICES, which torchada mirrors into CUDA_VISIBLE_DEVICES on import.
- Fall back from `adamw_torch_fused` to `adamw_torch` on MUSA, since `torch.optim.AdamW(fused=True)` is not supported there (transformers defaults to the fused variant on torch >= 2.8).
- Add `requirements/musa.txt` and document MUSA as supported hardware.

Not covered: the Megatron backend (requires megatron-core>=0.16 with MUSA patches), and the Web UI / Ray visible-devices handling.

Tested on MTT S5000 (torch_musa 2.9.1, torchada 0.1.89): unit tests in tests/utils/test_musa.py, plus LoRA SFT on Qwen2.5-0.5B-Instruct on one card and DDP on two cards, and inference with the trained adapter.

# PR type
- [ ] Bug Fix
- [x] New Feature
- [ ] Document Updates

# PR information

## Experiment results


swift/utils/torch_utils.py
```diff
@@ -12,13 +12,23 @@
...
 from .env import get_dist_setting, get_node_setting, is_dist, is_local_master, is_master, is_mp
 
 
+@lru_cache
+def is_torch_musa_available() -> bool:
+    """True when torch_musa (Moore Threads GPU) is loaded and a MUSA device is usable.
+
+    ``torch.cuda.is_available()`` is intentionally False on MUSA (see torchada), so MUSA needs its own check.
+    """
+    return getattr(torch, 'musa', None) is not None and torch.musa.is_available()
```
hjh0119: Can we follow the same pattern as other backends here
```
from transformers.utils import is_torch_musa_available
```

swift/__init__.py
```diff
@@ -1,7 +1,16 @@
...
+from .utils.import_utils import _LazyModule, is_torch_musa_installed, is_torchada_available
+
+if is_torch_musa_installed() and is_torchada_available():
+    # The visible devices must be final before torch is imported: torch autoloads torch_musa, which reads
+    # MUSA_VISIBLE_DEVICES only once.
+    from .cli.utils import sync_musa_visible_devices, try_use_single_device_mode
+    sync_musa_visible_devices()
+    try_use_single_device_mode()
+    # torchada redirects the torch.cuda.* APIs to torch.musa, so it must be imported before anything uses them.
+    import torchada  # noqa: F401
```
hjh0119: Why is this needed here?

caizhi-mt: ▎ torchada has to be imported before anything touches torch.cuda.* (it redirects those APIs to torch.musa), so it needs to sit at the package entry that every process goes through: CLI subcommands, torchrun workers and Python API users (from swift import ...). The visible-device lines before it matter too. import torch autoloads torch_musa, which reads MUSA_VISIBLE_DEVICES only once, so the env has to be final before torch is imported (verified on MUSA Platform). On other platforms the block is a no-op: it only calls find_spec and never imports torch. If you'd prefer a slimmer __init__.py, I can move it into a torch-free helper and keep a single call here.

swift/utils/torch_utils.py
caizhi-mt: Good point, switched to from transformers.utils import is_torch_musa_available. Note it was added in transformers 4.45.0, so I added transformers>=4.45 to requirements/musa.txt. Is it fine that importing it unguarded in torch_utils.py raises on transformers < 4.45, given framework.txt still says >=4.33? Otherwise I can guard the import.

hjh0119: Except for a very small number of limited-version models that are compatible, we mainly maintain transformers 5, so it's ok to add transformers>=4.45 to the requirement for musa.

swift/cli/utils.py
```diff
@@ -2,13 +2,25 @@
...
+    if 'MUSA_VISIBLE_DEVICES' not in os.environ and 'CUDA_VISIBLE_DEVICES' in os.environ:
+        os.environ['MUSA_VISIBLE_DEVICES'] = os.environ['CUDA_VISIBLE_DEVICES']
+
+
 def try_use_single_device_mode():
     if os.environ.get('SWIFT_SINGLE_DEVICE_MODE', '0') == '1':
-        visible_devices = os.environ.get('CUDA_VISIBLE_DEVICES')
+        # On MUSA the runtime reads MUSA_VISIBLE_DEVICES (see `sync_musa_visible_devices`).
+        env_key = 'MUSA_VISIBLE_DEVICES' if 'MUSA_VISIBLE_DEVICES' in os.environ else 'CUDA_VISIBLE_DEVICES'
+        visible_devices = os.environ.get(env_key)
```
hjh0119, line 37: Are there any cases where CUDA_VISIBLE_DEVICES would need to be set on MUSA?

hjh0119: maybe you can pop CUDA_VISIBLE_DEVICES in sync_musa_visible_devices

caizhi-mt: Good call. sync_musa_visible_devices now pops CUDA_VISIBLE_DEVICES after moving it to MUSA_VISIBLE_DEVICES, so on MUSA MUSA_VISIBLE_DEVICES is the only source of truth. I checked the remaining readers: the only one on the MUSA path was _set_visible_devices_for_dp_rank in rollout.py, which I switched to MUSA_VISIBLE_DEVICES on MUSA (it was already ineffective there, since workers inherit MUSA_VISIBLE_DEVICES).

swift/callbacks/activation_cpu_offload.py
```diff
@@ -15,6 +15,7 @@
 
 is_cuda_available = torch.cuda.is_available()
 is_npu_available = is_torch_npu_available()
+is_musa_available = is_torch_musa_available()
```
hjh0119, line 18: The comment says 'MUSA goes first: MUSA patches such as megatron-lm-musa-patch make torch.cuda.is_available() return True.' Given that, do we need to move this part up?

swift/utils/torch_utils.py
```diff
@@ -258,6 +274,8 @@ def ipc_collect():
         torch.cuda.ipc_collect()
     elif is_torch_npu_available():
         torch.npu.ipc_collect()
+    elif is_torch_musa_available():
+        torch.musa.ipc_collect()
```
hjh0119: Does this need to be moved up here as well?

swift/tuners/base.py
```diff
@@ -252,6 +252,8 @@ def load_state_file(path, device: Optional[str] = None):
                 device = 'cuda'
             elif is_torch_npu_available():
                 device = 'npu'
+            elif is_torch_musa_available():
+                device = 'musa'
```
hjh0119: Does this need to be moved up here as well?

swift/callbacks/activation_cpu_offload.py
caizhi-mt: Yes, moved MUSA first in get_device_name (534c952), matching get_torch_device

swift/utils/torch_utils.py
caizhi-mt: Yes, moved up as well in 534c952. Tested on MTT S5000: tests/utils/test_musa.py 36 passed.

swift/tuners/base.py
caizhi-mt: moved up as well in 534c952

docs/source/GetStarted/SWIFT-installation.md
```diff
@@ -161,6 +161,7 @@ modelscope-registry.us-west-1.cr.aliyuncs.com/modelscope-repo/modelscope:ubuntu2
 | RTX20/30/40系列 |                             |
 | T4/V100         | 部分模型出现NAN             |
 | Ascend NPU      | 部分模型出现NAN或算子不支持 |
+| 摩尔线程 MUSA   | 需安装 [torch_musa](https://github.com/MooreThreads/torch_musa)，并安装 [torchada](https://github.com/MooreThreads/torchada)（`pip install -r requirements/musa.txt`），swift 会自动导入 |
```
hjh0119, line 164: Could you please provide more details in the official documentation?

here are some examples:
- https://github.com/modelscope/ms-swift/blob/main/docs/source/BestPractices/NPU-support.md
- https://github.com/modelscope/ms-swift/blob/main/docs/source/BestPractices/Metax-support.md 
- https://github.com/modelscope/ms-swift/blob/main/docs/source/BestPractices/AMD-support.md

caizhi-mt: Sure, I'll add the details as soon as I can and get back to you.

